PR #37, opened 2022-05-20
Option for density and energy sinks for all species with sheath closure

johnomotani:
Allows source-driven 2d turbulence simulations.

If `sheath_closure:sinks = true` is set:
* Sheath particle flux is at a common sound speed of sqrt(P_total/(total mass density)), for all species. Electron particle sink is still modified by DivJsh.
* Ions have an energy sink like `gamma_i*Ti*sheath_flux`, where `gamma_i = 2.0` (by default) is taken from Stangeby's book.

NB These forms have not been chosen carefully - they were just the first that came to my mind!

### DIFF
--- a/docs/sphinx/examples.rst
+++ b/docs/sphinx/examples.rst
@@ -172,7 +172,7 @@ so only the electron density and vorticity are evolved. A sheath-connected
 closure is used for the parallel current.
 
 .. figure:: figs/blob2d.png
-   :name: blob2d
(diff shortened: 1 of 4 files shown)


johnomotani:
I have an input file for a 2d turbulence simulation - should I add that to the examples?

bendudson (approved):
Cool! Thanks @johnomotani 
An example would be great if that's ok

PR #59, opened 2022-09-22
Adding hydrogen CX diagnostics

bendudson:
Need to be a bit careful with naming of sources/sinks when there are multiple isotopes.

For two species `a` and `b`, the channel `Fab_cx` is a source of momentum for species `a` due to charge exchange with species `b`. There are corresponding sinks for the products of the charge exchange reaction which are not saved. 

For example: 
Reaction `d + t+ -> d+ + t` will save the following forces (momentum sources):
- `Fdt+_cx` is a source of momentum for deuterium atoms `d` and sink of momentum for deuterium ions `d+`.
- `Ft+d_cx` is a source of momentum for tritium ions `t+` and sink of momentum for tritium atoms `t`

The reason for this convention is the existence of the inverse reactions:
`t + d+ -> t+ + d` outputs diagnostics `Ftd+_cx` and `Fd+t_cx`.

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -691,6 +691,21 @@ twice.
 When reactions are added, all the species involved must be included, or an exception
 should be thrown.
 
+Notes:
+
+1. Charge exchange channel diagnostics: For two species `a` and `b`,
+   the channel `Fab_cx` is a source of momentum for species `a` due to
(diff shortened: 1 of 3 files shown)


mikekryjak:
I have tested the new variables against my Python CX code and I think they are correct. I think we can merge this unless you wanted to add something more?

PR #60, opened 2022-09-22
Some important fixes

bendudson:
1. Add charge exchange to collision frequency. This is important for neutral diffusion.
2. Correct copy/paste error in pressure correction for particle sources and flow velocity: `energy_source` was used rather than `density_source`.


mikekryjak (changes requested):
` R = Natom * Nion * sigmav;` is the rate coefficient which is equivalent to the CX reaction frequency, which is the same for both participating species. So should the collision frequency contribution not be as below instead?

```
  // Update collision frequency for the two colliding species
  add(atom1["collision_frequency"], Nion * Natom * sigmav);
  add(ion1["collision_frequency"], Nion * Natom * sigmav);
```

PR #65, opened 2022-10-07
Add neutral_boundary component

bendudson:
Adds a heat loss through Y boundaries for a single species.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ set(HERMES_SOURCES
     src/adas_reaction.cxx
     src/noflow_boundary.cxx
     src/neutral_parallel_diffusion.cxx
+    src/neutral_boundary.cxx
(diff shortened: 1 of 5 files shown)


mikekryjak:
@bendudson I get an error:
`====== Exception thrown ======
Option species:neutral_boundary:AA has no value`

This is odd as AA is definitely specified under the species component.
I tried adding another "AA" option under `[neutral_boundary]` but this didn't work.
Input file below:

[BOUT.txt](https://github.com/bendudson/hermes-3/files/9799039/BOUT.txt)

mikekryjak:
It would be good to expand this PR to cover the Franck-Condon energy issue as I think we should be implementing it in neutral_boundary. Happy to do this if I get unstuck on the AA issue somehow..
https://github.com/bendudson/hermes-3/issues/70

PR #67, opened 2022-10-10
Add diagnose option

mikekryjak:
If we are not expecting a perfect energy balance and don't save the sheath heat flux in the dump file, then there is no way for a user to validate their Python reconstruction of the sheath heat flux code.

This branch aims to save the heat flux for sheath_boundary, sheath_boundary_simple and neutral_boundary for all species involved.

Currently stuck on getting a segmentation fault after adding a diagnose option...


mikekryjak:
This one needs an update as there are now some sheath diagnostics present

PR #72, opened 2022-10-14
Add attributes to output

bendudson:
Adding attributes to output variables:
- `units` , a string containing the SI units (except eV for temperature)
- `conversion`, a float that the value should be multiplied by to convert to SI units
- `long_name`, a descriptive label
- `standard_name`, a short name that will hopefully be standard...

Also for some things:
- `species`, the name of the species the value belongs to
- `component`, the name of the component that calculated the value

Upgrading to latest BOUT++ `next`, replacing all `DataFile` use with `Options` to write data to output.


mikekryjak:
This is an excellent idea.

PR #75, opened 2022-10-24
Fix tokamak examples

bendudson:
It's been a while since these examples were tested
- Input file fixes, removing unused inputs
- evolve_density now requires charge and atomic mass to be set
- `anomalous_diffusion` was not setting Y boundaries, leading to NaN results. Not sure why these worked previously, but perhaps just not checked.

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -41,7 +41,20 @@ evolve_density
 ~~~~~~~~~~~~~~
 
 This component evolves the species density in time, using the BOUT++
-time integration solver.
+time integration solver. The species charge and atomic mass must be set,
+and the initial density should be specified in its own section:
+
+.. code-block:: ini
+
+   [d]
+   type = evolve_density, ...
+
+   AA = 2 # Atomic mass
+   charge = 0
+
+   [Nd]
+   function = 1 - 0.5x # Initial condition, normalised to Nnorm
+
 
 The implementation is in the `EvolveDensity` class:
 
--- a/examples/1D-hydrogen-equalT/BOUT.inp
+++ b/examples/1D-hydrogen-equalT/BOUT.inp
@@ -58,9 +58,10 @@ Tnorm = 100
 
 [solver]
 type = beuler  # Backward Euler steady-state solver
+snes_type = newtonls
 max_nonlinear_iterations = 10
 
-diagnose = true
+diagnose = false # Print diagnostic information?
 
 atol = 1e-7
 rtol = 1e-5
--- a/examples/1D-hydrogen/BOUT.inp
+++ b/examples/1D-hydrogen/BOUT.inp
@@ -57,9 +57,12 @@ Tnorm = 100
 
 [solver]
 type = beuler  # Backward Euler steady-state solver
+
+snes_type = newtonls
+
 max_nonlinear_iterations = 10
 
-diagnose = true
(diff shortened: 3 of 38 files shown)


mikekryjak:
Thanks @bendudson! 
I have tried each example and they all execute apart from diffusion-transport which also has an unused temperature boundary condition like diffusion-conduction. I have also tested all to completion apart from the really expensive ones (multispecies and turbulence).

mikekryjak:
Diffusion-transport fixed

johnomotani:
`examples/tokamak/recycling-dthene` has some unused options (I'll push a commit to fix those). It also fails unless the `json_database` directory is symlinked from the example directory. What is the expected workflow? Does it just need documenting in the README that the user should `ln -s` the `json_database` directory?

johnomotani:
Something's broken in the blob2d* examples. When running with multiple processes, vorticity and phi have weird stuff at the process boundaries - maybe some missing communication call? I tried adding a call to communicate `Vort` in `vorticity.cxx`, but that didn't fix the issue...

johnomotani:
`1D-hydrogen`, `1D-hydrogen-equalT`, `1D-neon`, and `solkit-comparison/single-temperature` fail for me with an error like
```
0.000e+00          1       8.75e-03     3.8    0.0    2.2   94.7   -0.6
|  Step 0 of 200. Elapsed 0:00:00.0 ETA 0:00:01.8Error encountered in solver run
Solver failed after many attempts
Finalising PETSc
Error encountered: Solver failed after many attempts====== Exception path ======
[bt] #5 ../../hermes-3(+0x356de) [0x561fded856de]
_start at ??:?
[bt] #4 /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x7f732e9ee083]
__libc_start_main at ??:?
[bt] #3 ../../hermes-3(+0x33c8f) [0x561fded83c8f]
main at ??:?
[bt] #2 /home/john/soltransport/hermes-3/build/external/BOUT-dev/lib/libbout++.so.5.0.0(_ZN6Solver5solveEid+0x822) [0x7f732f479ba2]
Solver::solve(int, double) at ??:?
[bt] #1 /home/john/soltransport/hermes-3/build/external/BOUT-dev/lib/libbout++.so.5.0.0(+0x140138) [0x7f732f1a8138]
SNESSolver::run() [clone .cold] at snes.cxx:?

Solver failed after many attempts

--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 3 in communicator MPI COMMUNICATOR 3 DUP FROM 0
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
[spike:69410] PMIX ERROR: UNREACHABLE in file ../../../src/server/pmix_server.c at line 2193
[spike:69410] 3 more processes have sent help message help-mpi-api.txt / mpi-abort
[spike:69410] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

johnomotani:
Thanks @bendudson - `examples/1D-hydrogen` runs for me now. It seems to be significantly faster in serial than in parallel - apparently the `ilu` preconditioner is much better than `bjacobi` or `hypre`... Maybe worth noting this somewhere?

PR #79, opened 2022-10-26
Add recycle_energy to recycling component

bendudson:
Gives recycled particles a non-zero energy, for example the Franck-Condon energy of recycled hydrogen atoms.


mikekryjak:
Thanks @bendudson, this is perfect - a really intuitive way to implement this.

PR #87, opened 2022-11-24
Rename `loadmetric` to `recalculate_metric`. Set `calcParallelSlices` to false by default if not 3D metrics.

bendudson:
Rename `loadmetric` option to `recalculate_metric` and change the default from `true` to  `false`. Hypnotoad saves metrics to the grid file so we usually don't need to recalculate. See issue #286 

If metrics are 2D, Hermes-3 operators don't need parallel slices, so don't calculate them by default.


mikekryjak:
Looks good to me. But since most simulations in the wild will have a `loadmetric = false` line in the input file, think we should:

1. Send a mass email to the users notifying them to expect an error when they update Hermes-3 and the reason for it, or
2. Catch the old setting and print a warning about it.

I would vote for 1) to keep things simple since we have the email address of practically every user.

bshanahan:
I think BOUT++ throws an error if there are unused inputs (which would be the case now with `loadmetric`. So, this error-catching should already exist.

mikekryjak:
@bshanahan that's a very good point. So maybe we can merge it as is?
Turns out there are actually two issues that this solves: https://github.com/bendudson/hermes-3/issues/212 and https://github.com/bendudson/hermes-3/pull/286 which has already been mentioned.

bshanahan:
@bendudson we found difficulty getting Zoidberg J calculations to be smooth. As a result  we often neglect J in the grid file, and rely on the internal calculation within BOUT++, which spits a warning:
` WARNING: Jacobian 'J' not found. Calculating from metric tensor`

Is this an incorrect way of doing this? The most recent change will throw an exception in these cases.

bendudson:
Hey @bshanahan Good point! I think the approach you're using is fine. The only case I'm concerned about is old grid files that only have `Rxy`, `Bpxy` etc. Without the `J` check this PR would silently break those simulations. What about checking for at least one metric component: Only throw an exception if none of (J, g11, g_11, g22, g_22 etc.) are in the grid file. That should accept your stellarator grids without J, but throw an exception for old grid files with no metrics.

PR #88, opened 2022-11-29
WIP: Operators for non-orthogonal meshes

bendudson:
Including off-diagonal metric terms coming from non-orthogonal meshes. Starting with operators in neutral fluid model as those are likely to have the most impact.


dschwoerer:
I have written a small MMS test and have been running it for the div_a_grad_perp_nonorthog, but it does not seem to converge:

dagpno

For reference, the Delp2 operator:
delp2

I am trying to check zoidberg is correct, so the bug might be in zoidberg, but as Delp2 passes, I thought I mention it ...

The test converges if the grid is orthogonal, but fails otherwise.

mikekryjak:
@dschwoerer Michael Hardman started with us and will be taking a look at this, do you mind sharing your MMS test?

bendudson:
Replaced by #285

PR #94, opened 2022-12-07
Extra diagnostics and settings for anomalous diffusion

mikekryjak:
It is common to run with a profile of anomalous diffusion coefficients (e.g. with a lower value in the core to represent the transport barrier). It is also useful to change the coefficients while running the model, for example by only imposing a transport barrier after steady state has been achieved with uniform coefficients.

Unfortunately due to input file limitations restricting Y to be continuous around the core, it is not possible to differentiate the PFR from the core in the input file using Heaviside functions. This means that a coefficient profile must be imported from the mesh file (see [PR](https://github.com/bendudson/hermes-3/pull/86)). This poses two problems:

- It is not possible to check that the code is reading the coefficients correctly due to lack of diagnostics
- It is not possible to modify the coefficients in the input file, they must be changed in the mesh file.

In an ideal world:

- The mesh would contain only the final coefficients and one could scale them from the input file
- The output file would contain the coefficient settings, allowing one to track their changes over time and verify the mesh file is correct

This PR aims to do the following:

- [x] Add diagnostics for the anomalous coefficient settings
~- [ ] Add new input settings to the input which scale the coefficients~

**Update 20/01/2023:**
- Anomalous D, Chi and Nu now saved as new diagnostic variables
- Abandoned idea of scaling the sources from file. This is because in setting up a transport barrier, one would want to gradually reduce the coefficients in the core specifically. Can't really think of many uses for scaling the entire profile now, and one can already scale it up and down using the input file if one wants the coefficients to be uniform.

### DIFF
--- a/include/anomalous_diffusion.hxx
+++ b/include/anomalous_diffusion.hxx
@@ -22,7 +22,11 @@ struct AnomalousDiffusion : public Component {
   ///   - anomalous_nu   Overrides nu_<name>
   ///   - anomalous_sheath_flux  Allow anomalous flux into sheath?
   //                             Default false.
-  AnomalousDiffusion(std::string name, Options &alloptions, Solver *);
+  AnomalousDiffusion(std::string name, Options &alloptions, Solver *) {
+    diagnose = alloptions[name]["diagnose"]
+                   .doc("Output additional diagnostics?")
+                   .withDefault<bool>(false);
+  };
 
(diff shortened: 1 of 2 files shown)


mikekryjak:
@bendudson this one is ready for merge I think. Can you review?

bendudson (approved):

PR #101, opened 2023-01-18
Low density floors

bendudson:
- Apply non-zero floors to density when dividing in `zero_current` and `electron_force_balance`
- Add check in function to set values in the state, to catch non-finite values early

### DIFF
--- a/include/component.hxx
+++ b/include/component.hxx
@@ -200,6 +200,25 @@ T getNoBoundary(const Options& option, const std::string& location = "") {
   getNoBoundary<Type>(option)
 #endif
 
+/// Check whether value is valid, returning true
+/// if invalid i.e contains non-finite values
+template<typename T>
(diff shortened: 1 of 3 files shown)


include/component.hxx
```diff
@@ -200,6 +200,25 @@ T getNoBoundary(const Options& option, const std::string& location = "") {
...
+/// Check whether value is valid, returning true
+/// if invalid i.e contains non-finite values
+template<typename T>
+bool hermesDataInvalid(const T& value) {
+  return false; // Default
+}
+
+/// Check Field3D values.
+/// Doesn't check boundary cells
+template<>
```
dschwoerer, line 212: Quick question: Why do you use a template here rather than a inline function?

bendudson: I think because it's called from the `set` function, which itself is a template. It only has an implementation for `Field3D` but may be instantiated for other types.

PR #105, opened 2023-02-07
Neutral full velocity model

bendudson:
Solves Navier-Stokes equations for axisymmetric neutral gas: All components of velocity are evolved as Field2D quantities (i.e dependent on X and Y but not Z).

Manual page describes the model and its settings.

Includes PR #350 because the `softFloor` function speeds up this model too.

By default momentum advection (dynamic pressure) is neglected, so that the steady state flow results from the same balance of static pressure and friction as `neutral_mixed`. Switches turn on momentum advection, and the centrifugal force / angular momentum conservation terms due to curved toroidal domain.


mikekryjak:
@bendudson Love to see some activity on this PR! What made you return to it and what's the plan?

PR #119, opened 2023-04-12
Neutral model perpendicular BC improvements

mikekryjak:
We need the ability to easily define neutral transport at the boundaries. In particular, need an intuitive way to allow neutral density and momentum to escape the core and neutral pressure to escape the SOL. Right now this is complicated because there are hardcoded BCs on both neutral parameters as well as diffusivities (see https://github.com/bendudson/hermes-3/issues/117).

- [x] Expose neutral parameter BCs in the input file
- [x] Expose neutral diffusivity BCs in the input file
- [x] Change neutral BC defaults so that they are Neumann if no user setting is changed (to match what was hardcoded before)


src/neutral_mixed.cxx
```diff
@@ -31,6 +31,7 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
   solver->add(Nn, std::string("N") + name);
   solver->add(Pn, std::string("P") + name);
   solver->add(NVn, std::string("NV") + name);
+  /// solver->add(Dnn, std::string("Dnn" + name));
```
bendudson, line 34:
```suggestion
```

bendudson (approved):

src/neutral_mixed.cxx
mikekryjak: Thanks @bendudson Not sure why I left that there, or why Dnnd didn't need to be treated this way. I have a feeling that I tried to add it to match the other variables, but then found that it made Hermes-3 crash. What does `solver->add()` do?

bendudson: Hey @mikekryjak `solver->add()` adds the variable to the time integrator (e.g. CVODE), making it a time-evolving quantity. The solver will set the variable value, and expect `ddt()` of that variable to be set.

mikekryjak:
Ignore the previous comment - I got myself confused. I changed the defaults for the neutral diffusion, not the anomalous diffusion. I think that means there's a bug.

PR #121, opened 2023-04-28
WIP: PI Controller fixes

mikekryjak:
PR for branch originally started by @bendudson to address this issue:
https://github.com/bendudson/hermes-3/issues/82

Ben's original comment:
Separating density source from upstream feedback
source now specifies a fixed (not time varying) source, whicle source_shape is used by the upstream_density_feedback component.

This has technically enabled the PI controller to work, but there are still problems, see comments on the issue. 

Update 25/05/2023

- [x] Fix PI controller
- [x] Add diagnostic variables
- [x] Tune example cases
- [x] Update tests
- [x] Update documentation


mikekryjak:
I've done a bunch of tests and it looks like changing the P/I parameters does affect the controller, but it's just that the I component is so large that it dominates by several orders of magnitude even with the default settings where the I coefficient is 2x smaller than the P coefficient.

So in short, the controller appears functional, but the default settings are not good and lead to an overwhelmingly I dominated source (for whatever reason) while the overall source is too small to have any effect (this is because the source_shape was missing a multiplier in [m-3 s-1] in the input file).

I'm happy to try and tune P/I to get a good result, but I am unable to add the proportional and integral source components to saved variables for diagnostic purposes.
Xarray complains that `conflicting values for variable <feedback variable> on objects to be combined`. This is very odd, but I suspect it's something to do with the error calculation. I wonder if this is why there is a mysterious line which broadcasts the multiplier from the first processor (https://github.com/bendudson/hermes-3/blob/b8308c671fc4ecdeb574a1f459db01dd409bf412/src/upstream_density_feedback.cxx#L49) @bendudson what does this mean, why is it needed?

mikekryjak:
@bendudson this is now ready for merge after the `pull_request` test is fixed. I have no idea why it's failing - the `push` test is showing the expected result, but the `pull_request` test outcome makes no sense: it claims that 145eV is greater than the 150eV in the test script. One possible issue is that the `pull_request` test is referencing a different script that I haven't seen.

bendudson (approved):
Thanks @mikekryjak !

PR #128, opened 2023-05-19
WIP: Adding a `radas` component

bendudson:
Will read radiation cooling curves from RADAS:
https://github.com/cfs-energy/radas


bendudson:
Probably not needed yet.

PR #129, opened 2023-05-19
Upgrade cooling curves for fixed fraction impurity radiation

mikekryjak:
The current cooling curves for N, Ar and Ne have a limited temperature range and have discontinuities which are contributing to significant slowdowns in deep detachment. The slowdowns manifest themselves as spikes when the plasma is perturbed, and in particular when the detachment front is moving from one grid to another as it travels to the X-point.

The new curves will come from ADAS and will be continuous fits. I have currently implemented a curve provided by David Moulton (originating from ADAS with tau = 0.5ms). The fit is a continuous 10 coefficient polynomial valid between 0.2 and 2000eV, see picture.  I will install [radas](https://github.com/cfs-energy/radas) and generate curves myself later.

Update 05/06/2023: radas curves for Ne, Ar and N are implemented, see pictures below. Testing in progress
Update 04/07/2023: Implemented C, improved precision in coefficients and updated documentation. Fit error is within 2% in the test case and is due to coefficient rounding error.

- [x] Implement curve from David
- [x] Test
- [x] Generate radas curves for Ne, Ar and N and implement them
- [x] Test
~~- [ ] Possibly include a way to select more than one curve? Needs discussion due to how the component is structured~~
- [x] Carbon

Final fits are below. The fit was chosen as a compromise between low temperature, high temperature and peak accuracy while maintaining the 10 coefficient fit format. The errors are easily within the error bounds of the fixed fraction assumption.
argon_fit
carbon_fit
neon_fit
nitrogen_fit


mikekryjak:
The latest rates seem to be off compared to what was intended to be implemented by 0% - 15% depending on the species. Currently testing to see what it could be.

mikekryjak:
Issue with match was purely due to insufficient precision (4 decimal points). Adding 8 decimal points reduces the max fit error to within 2% which I think is probably sufficient.

bendudson (approved):
Thanks @mikekryjak !

dschwoerer:
@mikekryjak do you mind sharing the notebooks you used to generate the code?

mikekryjak:
@dschwoerer sure. In fact, I think they could use being uploaded to the repo so that we can reuse the scripts to generate curves in the future. I'll do a PR to that effect

mikekryjak:
A note on the difference between the new and old curves. Ar, Ne and N were previously done in 0.1ms x 1e20m-3 and the new residence time of 0.5ms results in increased radiation in the peak region, see below for Ar where the old curve is labelled as DLS. 
image

The old carbon curve came from a paper by Hutchinson where it is not referenced. It seems like a very simple fit. The new curve is from ADAS and has substantially less radiation. This is a bit puzzling as Hutchinson's rate is meant to be in the coronal approximation and should therefore have less radiation. The Hutchinson rate is preserved in the `fixed_fraction_hutchinson_carbon` component. The old Ar, Ne and N rates are not preserved due to their impact on solver performance.
image

Philbert1999:
@mikekryjak Hi! I have installed [radas](https://github.com/cfs-energy/radas) and generated some curves myself. I find that radas could plot both "noncoronal_electron_emission_prefactor" and "coronal_electron_emission_prefactor". For noncoronal prefactor, it varies in different n*tau, so radas plots a bunch of curves. As for coronal prefactor radas only plots one curve for one type of impurity. So why did you choose noncoronal prefactor, what if i use coronal one?

mikekryjak:
Hi @Philbert1999! The "coronal" in the coronal approximation refers to the assumption of very low density, which is equivalent to n*tau being in its lower limit. This is not a very good assumption in the scrape-off layer where the density and residence time has an impact. Unfortunately n*tau is not a constant number. When you simulate impurity transport then this is explicitly predicted by the simulation, but when using a fixed fraction model like this one you need to make an assumption. I took mine from discussions with David Moulton who found that 1e20 * 0.5ms was good in his work on MAST-U: https://iopscience.iop.org/article/10.1088/1361-6587/aa6b13/meta

Philbert1999:
@mikekryjak Thank you very much for your patient response! I have one final question. When I use the "fixed_fraction_radiation" component, I'm a bit confused about how to control the density distribution of a particular impurity. For example, if the impurity is injected from a certain direction, how do I specify its distribution? As per my understanding, impurities under this component seem to be uniformly distributed throughout the device, similar to electrons. Perhaps I have misunderstood something. Thank you!

mikekryjak:
Hi @Philbert1999. The fixed fraction model assumes that the density of the impurities is always a fixed fraction of the electron density, so the impurities do not have a separate distribution. This is a very simple way to represent impurity radiation, but it has a lot of benefits: it has almost no additional computational cost and captures a lot of the effects. 

If you want to evolve impurity transport, you can use this example as a reference: https://github.com/bendudson/hermes-3/tree/master/examples/1D-neon

In general this makes the simulation substantially slower and more unstable, because you must simulate the evolution of many new species (all the ionisation states of the impurity) and these may have very low pressures/densities in parts of the domain. This is the case for other edge codes as well. 
Because of the extra difficulty you may need to start with fixed fraction as the initial state.

P.S. We have a Slack user group channel, you're welcome to join and we could continue the discussion there! Send me an email at mike.kryjak@york.ac.uk and I can add you :)

Philbert1999:
@mikekryjak Once again thanks for your assistance! I have sent you an email :)

PR #140, opened 2023-06-07
Disable debugging

nass13m:


bendudson:
Thanks @nass13m ! For good performance by default this is probably the right thing to do.

PR #142, opened 2023-06-09
WIP: Add time-dependent sources

bendudson:
Implemented for density first, for discussion. There are several ways this can be done; so input on use-cases and the "least surprising" approach would be helpful.

The present implementation wouldn't allow a time-dependent source that added to or modified a source in the mesh input file, for example. Is that a useful thing to do?

If `source_time_dependent` switch is set, then the input source is re-evaluated every iteration. The source must be in the input options, and any value in the mesh file is ignored.

```
[e]
type = evolve_density

[Ne]
source_time_dependent = true
source = x * (1-x) * t  # Can depend on 't' in seconds
```

Note: The time 't' is in seconds, and the source should be in SI units i.e. m^-3/s

### DIFF
--- a/include/evolve_density.hxx
+++ b/include/evolve_density.hxx
@@ -80,6 +80,15 @@ private:
   Field3D source; ///< External input source
   Field3D Sn; ///< Total density source
 
+  bool source_only_in_core;  ///< Zero source where Y is non-periodic?
+  bool source_time_dependent; ///< Is the input source time dependent?
+  BoutReal source_normalisation; ///< Normalisation factor [m^-3/s]
+  BoutReal time_normalisation; ///< Normalisation factor [s]
+  FieldGeneratorPtr source_generator;
+
+  /// Modifies the `source` member variable
+  void updateSource(BoutReal time);
+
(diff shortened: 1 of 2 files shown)


mikekryjak:
Hi @bendudson, thanks a lot for this! I think there is a use case which requires the ability to define sources from the grid.
In 2D, if not using a Dirichlet core, one must specify a volumetric power source in the first core ring. This cannot be done in the input file because it does not differentiate between the PFR and the core. So if we have to switch to defining it only in the input file in order to issue a transient, we would not be able to reproduce the original steady state.
However, after going through the commits I see that you prepared a new input flag `source_only_in_core`. Is this specifically to address the above limitation?

The other grid depended sources would be any puffing I guess. Simulating a puff transient could be of interest for the purposes of detachment control, in which case having the ability to scale a source defined in the grid file would be useful. But I am not aware of anyone wanting to do this right now, so maybe we can implement it when it's needed?

mikekryjak:
Will test and extend to pressure source. Will ignore the issue with mesh sources and pick up as separate PR if and when required

mikekryjak:
@bendudson ready for review

PR #150, opened 2023-07-13
Add component for stellarator two-point model 

bshanahan:
This adds a component which adds binormal transport terms from the stellarator two point model:
http://dx.doi.org/10.1088/0741-3335/53/2/024009

Here we include the field-line pitch effects in the density equation (not just pressure and momentum as in the cited source). There are also additional terms for cross-field drifts in the density  and pressure equations.
 
There is an example provided.

### DIFF
--- a/include/binormal_stpm.hxx
+++ b/include/binormal_stpm.hxx
@@ -8,7 +8,7 @@
 
 /// Adds terms to Density, Pressure and Momentum equations following the
 /// stellarator 2-point model from Yuhe Feng et al., PPCF 53 (2011) 024009
-/// The terms add the effective parallel contribution of perpendicular drifts
+/// The terms add the effective parallel contribution of perpendicular transport
 /// in long connection length scenarios.
 /// B Shanahan 2023 <brendan.shanahan@ipp.mpg.de>
 struct BinormalSTPM : public Component {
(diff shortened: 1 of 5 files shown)


include/binormal_stpm.hxx
```diff
@@ -0,0 +1,49 @@
...
+#define BINORMAL_STPM_H
+
+#include <bout/field3d.hxx>
+
+#include "component.hxx"
+
+/// Adds terms to Density, Pressure and Momentum equations following the
+/// stellarator 2-point model from Yuhe Feng et al., PPCF 53 (2011) 024009
+/// The terms add the effective parallel contribution of perpendicular drifts
+/// in long connection length scenarios.
```
dschwoerer, line 12: It is always present - but likely negligible for tokamaks
```suggestion
/// which is of significance in long connection length scenarios.
```

include/binormal_stpm.hxx
```diff
@@ -0,0 +1,49 @@
...
+#ifndef BINORMAL_STPM_H
+#define BINORMAL_STPM_H
+
+#include <bout/field3d.hxx>
+
+#include "component.hxx"
+
+/// Adds terms to Density, Pressure and Momentum equations following the
+/// stellarator 2-point model from Yuhe Feng et al., PPCF 53 (2011) 024009
+/// The terms add the effective parallel contribution of perpendicular drifts
```
dschwoerer, line 11: It seems to includes advection and diffusion
```suggestion
/// The terms add the effective parallel contribution of perpendicular transport
```

bendudson (approved):
Thanks @bshanahan !

PR #151, opened 2023-07-14
Add notebook on new impurity curves

mikekryjak:
This PR adds a notebook documenting the process of implementing new cooling curves in:
https://github.com/bendudson/hermes-3/pull/129

The process can be replicated to change the fit or to implement new curves. The fitting routine prints out C++ code for the fit coefficients which you can copy paste directly into Hermes-3.

I also include a comparison between the old and the new rates as well as a validation plot.


bendudson (approved):
Thanks @mike! This looks great

PR #152, opened 2023-07-19
WIP: Cell face flow information

bendudson:
The `anomalous_diffusion` component now saves flows of particles, momentum and energy through cell faces in the state. The plan is for this to be extended to other components, making cell face flows easier to diagnose.

`PFx = state["species"]["e"]["particle_flow_xlow"]` is the (normalised) number of particles per second passing through the lower X cell face: `PFx(i,j,k)` is the number of particles per second passing from cell `(i-1,j,k)` to `(i,j,k)`. Note that positive flow means flow in the positive coordinate direction. 

Similarly, `PFy = state["species"]["e"]["particle_flow_ylow"]` is the flow of particles through the lower Y cell face.

The rate of change of density in cell `(i,j,k)` due to the flows is:
```
ddt(Ne) = ( PFx(i,j,k) + PFy(i,j,k) - PFx(i+1, j, k) - PFy(i,j+1,k) ) / dV
```
where `dV = J * dx * dy * dz` is the volume of the cell.

### DIFF
--- a/docs/sphinx/code_structure.rst
+++ b/docs/sphinx/code_structure.rst
@@ -64,12 +64,12 @@ for new variables which are added to the state:
     * `momentum_source` Normalised momentum source
     * `energy_source`  Normalised energy source
 
-    * `density_flow_xlow`  Normalised particle flow through lower X cell face
-    * `density_flow_ylow`  Normalised particle flow through lower Y cell face
+    * `particle_flow_xlow` Normalised particle flow through lower X cell face
+    * `particle_flow_ylow` Normalised particle flow through lower Y cell face
     * `momentum_flow_xlow` Normalised momentum flow through lower X cell face
     * `momentum_flow_ylow` Normalised momentum flow through lower Y cell face
-    * `energy_flow_xlow`  Normalised energy flow through lower X cell face
-    * `energy_flow_ylow`  Normalised energy flow through lower Y cell face
+    * `energy_flow_xlow`   Normalised energy flow through lower X cell face
+    * `energy_flow_ylow`   Normalised energy flow through lower Y cell face
 
 * `fields`
 
--- a/docs/sphinx/code_structure.rst
+++ b/docs/sphinx/code_structure.rst
@@ -64,12 +64,12 @@ for new variables which are added to the state:
     * `momentum_source` Normalised momentum source
     * `energy_source`  Normalised energy source
 
-    * `density_flux_xlow`  Normalised particle flux through lower X cell face
-    * `density_flux_ylow`  Normalised particle flux through lower Y cell face
-    * `momentum_flux_xlow` Normalised momentum flux through lower X cell face
-    * `momentum_flux_ylow` Normalised momentum flux through lower Y cell face
-    * `energy_flux_xlow`  Normalised energy flux through lower X cell face
-    * `energy_flux_ylow`  Normalised energy flux through lower Y cell face
+    * `density_flow_xlow`  Normalised particle flow through lower X cell face
+    * `density_flow_ylow`  Normalised particle flow through lower Y cell face
+    * `momentum_flow_xlow` Normalised momentum flow through lower X cell face
+    * `momentum_flow_ylow` Normalised momentum flow through lower Y cell face
+    * `energy_flow_xlow`  Normalised energy flow through lower X cell face
+    * `energy_flow_ylow`  Normalised energy flow through lower Y cell face
 
 * `fields`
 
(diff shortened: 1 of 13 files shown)


mikekryjak:
Amazing, thank you @bendudson!!

mikekryjak:
@bendudson 

When I run with the new PR I get a crash: 

Option :particle_flow_xlow does not exist

I think I finally worked out why - the flows are saved in `evolve_energy`, `evolve_momentum` and `evolve_density` but these are loaded before `anomalous_diffusion` where the flows are actually calculated. I don't think we can run `anomalous_diffusion` ahead of those three, although correct me if that is wrong.

Also, am I correct in thinking that `*_flow_ylow` is only the poloidal component of perpendicular diffusion at the moment? Is your intent to add the other flows to this and keep the naming convention? I think it would be nice to save the individual components of the transport if we can, but maybe we can think about splitting it later once everything works.

bendudson:
Sorry @mikekryjak ! I should have tested it first. Silly mistake in the `evolve_*` components. Now fixed and running.

mikekryjak:
Thanks @bendudson, I didn't realise that, I should've looked at the structure of the `state.` 
Was I right in thinking that the `ylow` component at the moment is just the poloidal component of perpendicular anomalous diffusion? Are you wanting to sum the other poloidal flows into that at a later time?

bendudson:
Hey @mikekryjak Yes that's correct: At the moment only `anomalous_diffusion` sets these flow quantities, but in future I'm imagining other components will add their flows. This way a boundary condition that uses the flows can take into account all processes that lead to flows: If there are drifts causing flows into the boundary, in addition to diffusion, then that should be part of the flow. It would be nice to have the option to separate components for diagnostics too.

mikekryjak:
@bendudson how much additional development did you intend this PR to have? From my end, I would just need the radial neutral transport and don't need anything more for now. This is so that we can extend `neutral_boundary.cxx` to the SOL/PFR as per https://github.com/bendudson/hermes-3/issues/160.

I think saving all the different poloidal fluxes is probably going to be more difficult, so maybe we can leave that till another PR? 
If we do that, maybe it's better to not save the poloidal fluxes we have at the moment as they are only the poloidal projection of the perpendicular anomalous transport, even though they are named as if they were total poloidal flows. Or maybe we can rename them?

bendudson:
Thanks @mikekryjak ! I think just getting what we need in there is probably enough for now. Adding all the other possible radial or poloidal fluxes is indeed a lot more work. We should document that the flux is incomplete; perhaps in the long name description here https://github.com/bendudson/hermes-3/pull/152/files#diff-04a800fe4c5382928cd32e41b555e68f384c5ba08082ff4f9c05c4e78a401b43R315
or somewhere in the manual?

mikekryjak:
@bendudson sounds good. Is this something you can do or shall I try and modify the neutral diffusion operators?

bendudson:
Hey @mikekryjak  I've added a warning note to the output diagnostics. Hopefully someone will get around to adding other terms to these fluxes, but we can't guarantee that they're complete.

bendudson:
Hi @mikekryjak The neutral fluxes are proving interesting. The `neutral_mixed` component calculates cross-field diffusion fluxes in the `finally` function using the collision frequency. That calculation can't be put into `transform` because the collision frequency isn't known at that point. After `finally` calculates the fluxes they can't be put into the state for another component to use.

Possible solutions are:
1. Add the neutral boundary conditions to `neutral_mixed`. This is probably the quickest to implement.
2.  Remove `neutral_mixed` and replace it with a `neutral_diffusion` top-level component, that would use the collision frequencies to calculate cross-field diffusion fluxes. Neutral particles would use `evolve_density`, `evolve_momentum` and `evolve_pressure` components the same as charged particles. This is a more significant change, but might be cleaner in the long run.

I suggest going for (1) for now. Once we are happy(ish) with the boundary conditions on neutral particle fluxes, we can refactor it as in (2).

mikekryjak:
Hi @bendudson. Roger that, I'll put them into `neutral_mixed`. This will be thermal reflection for both the edge and the targets. 
It would be good to talk through `neutral_boundary` in https://github.com/bendudson/hermes-3/issues/160 and make sure I'm not missing anything. 
I am also completely in favour of getting rid of `neutral_mixed` and refactoring it to be more consistent with the rest of the code. I think it would make things less confusing for new users.

mikekryjak:
@bendudson the previous pictures were with a poor choice of colormap and did not show the sign of the flows.
It looks like flow in the positive coordinate direction is negative, which is the opposite to what was intended (I think?)
image

mikekryjak:
@bendudson I flipped the signs. I think this one is good to go. 
The change somehow also made the artifacts around the X-point disappear - compare the picture below to the previous post.

image

PR #155, opened 2023-07-21
Add edge recycling

mikekryjak:
First of several BCs required to bring Hermes-3 in line with UEDGE/SOLPS/SOLEDGE2D/etc in terms of boundary conditions.
This requires the PR which enables cell face flows to be saved in the state: https://github.com/bendudson/hermes-3/pull/152
It's required because the perpendicular fluxes are currently calculated in `anomalous_diffusion` and not saved; the operators only output the change in source to the next cell as a result of the flux. At the moment, just the radial and poloidal components of the three anomalous transport channels are saved. These are sufficient to calculate the plasma flux leaving the SOL edge.

Note that for plasma flux to leave the edge you need a free boundary condition on both density (to allow a density gradient and therefore perpendicular diffusion) as well as pressure (to avoid violating P=NT) but a neumann condition on temperature (to prevent conduction). This requirement will need documentation and maybe a warning if the user only set a neumann condition. 

One of the other things I will be implementing is a decay length boundary condition which will be an alternative to the free BC in the above scenario.

Tasks:
- [x] Add recycling in the SOL edge
- [x] Add recycling in the PFR edge
- [x] Add diagnostic variables for particle and energy sources from recycling
- [ ] ~Implement console warning if user sets edge recycling but fails to set free boundary condition~
- [x] Make BC requirements clear in the documentation
- [x] Add individual controls for SOL/PFR/target recycling fraction and recycling species energy
- [x] Simple tests
- [x] Complex test
- [x] Mass balance test
- [x] Update examples
- [x] Add documentation

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -978,17 +978,17 @@ The boundary fluxes might be set by sheath boundary conditions,
 which potentially depend on the density and temperature of all species.
 Recycling therefore can't be calculated until all species boundary conditions
 have been set. It is therefore expected that this component is a top-level
-component which comes after boundary conditions are set.
+component (i.e. in the `Hermes` section) which comes after boundary conditions are set.
 
 Recycling has been implemented at the target, the SOL edge and the PFR edge.
-Each is off by default and must activated with a separate flag. Each can be 
+Each is off by default and must be activated with a separate flag. Each can be 
 assigned a separate recycle multiplier and recycle energy. 
 
 The chosen species must feature an outflow through the boundary - any cells
 with an inflow have their recycling source set to zero. If a sheath boundary condition
 is enabled, then this is automatically satisfied at the target through the Bohm condition.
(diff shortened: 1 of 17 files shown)


mikekryjak:
@bendudson this one is ready to go.
Note that it breaks existing input files: you have to change "recycle_multiplier" to "target_recycle_multiplier". I've updated the examples.

mikekryjak:
@bendudson I need to do one more check, I just remembered I was getting an unexpected segmentation fault on one of the more recent commits on this one. I am doing another full res case test on the latest commit with the master merged

mikekryjak:
@bendudson this is now ready to go. My crash was because I was asking for PFR recycling while my PFR BC was still Neumann due to a typo (BOUT++ wants `bndry_pf` instead of `bndry_pfr`, I keep mixing them up).

bendudson (approved):

PR #163, opened 2023-08-02
WIP: Add neutral pump

mikekryjak:
Adds neutral pump for 2D/3D cases. This allows a more physically realistic setup: the target recycling is set to 1 (or almost 1) and the domain particle balance is then controlled by the pump recycle multiplier (i.e. the pump albedo). This means the particle outflow is no longer controlled by the amount of ions flowing into the target along the parallel, but instead by the amount of neutrals flowing to the pump in the perpendicular direction. This kind of condition is common in other codes such as UEDGE, SOLPS and SOLEDGE2D.

The pump location is specified in the grid file in line with what other codes do. You must pass a new Field3D called `is_pump` in the grid file and select pump cells by setting them to `1`. Only cells in the PFR and SOL edge will be picked up - careful not to set it in the guard cells.

The pump can be enabled or disabled by passing `neutral_pump` and its albedo controlled by `pump_recycle_multiplier` like the other recycling sources. You **must** have recycling enabled on the boundary edge where the pump resides.

~The code implementation is quite simple: the `recycling.cxx` component already iterates through the SOL and PFR boundaries if recycling is enabled there. If Hermes-3 finds a cell with `is_pump = 1`, it swaps the `recycle_multiplier` of the boundary with the `pump_recycle_multiplier`. The recycling sources from the pump are put into new diagnostic variables `Sx_pump_recycle` and `Ex_pump_recycle` and the diagnostic variables for sources from the host boundary will no longer count the pump flows.~

The implementation is tricky and needs discussion.

Tasks:

- [x] Implementation of pump taking out incident ions
- [x] Implementation of pump taking out incident atoms
- [x] Simple test
- [x] Complex test
- [x] Particle/energy balance test
- [x] Add documentation

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -1189,9 +1189,55 @@ Each returning atom has an energy of 3.5eV:
    sol_recycle_multiplier = 1 # Recycling fraction
    sol_recycle_energy = 3.5   # Energy of recycled particles [eV]
 
-   sol_recycle = true
-   sol_recycle_multiplier = 1 # Recycling fraction
-   sol_recycle_energy = 3.5   # Energy of recycled particles [eV]
+   pfr_recycle = true
+   pfr_recycle_multiplier = 1 # Recycling fraction
+   pfr_recycle_energy = 3.5   # Energy of recycled particles [eV]
+
+Neutral pump
+^^^^^^^^^^^^^^^
+
+The recycling component also features a neutral pump, which is currently implemented for 
+the SOL and PFR edges only, and so is not available in 1D. The pump is a region of the wall
+which facilitates particle loss by incomplete recycling and neutral absorption.
+
+The particle loss rate :math:`\Gamma_{N_{n}}` is the sum of the incident ions that are not recycled and the 
+incident neutrals which are not reflected, both of which are controlled by the pump multiplier :math:`M_{p}` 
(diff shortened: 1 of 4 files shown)


mikekryjak:
I made a big omission in figuring out how to implement this. I thought of the pump as a surface with a low recycling coefficient, but it's primarily a surface that absorbs neutrals! The recycling bit is implemented and works, but the neutral absorbing bit is more complicated. I think we would need to change the boundary condition in the wall where the pump is to allow particle diffusion (free BC on pressure and density). Then we would need to make sure that `neutral_mixed.cxx` actually picks this up and the transport happens. And then we need to get the resulting transport and implement a pump neutral source since not all neutrals will be absorbed!

I think we need to put in quite a bit of thought into this - the way the neutral model is split between `neutral_mixed`, `recycling` and `neutral_boundary` is already pretty confusing.

mikekryjak:
After a discussion it seems that there's a simple way after all. We can assume that the thermal velocity is higher than the perpendicular velocity, just like for the wall reflective cooling model in `neutral_boundary`. This way we don't rely on the cross-field transport operator to calculate our fluxes. We can use the approach from `sheath_boundary` to extrapolate neutral density and temperature to the wall in the BC while maintaining a `neumann` boundary condition in the code. We then calculate thermal velocity at the wall and this way we get the incident particle and heat flux.

mikekryjak:
@bendudson ready for review

bendudson (approved):
Thanks @mikekryjak !

PR #168, opened 2023-08-14
Add collision diagnostics

mikekryjak:
This PR is now scaled back to just the collision frequencies due to a change in priorities. The other objectives will be picked up by a separate PR.
Tasks:

- [x] Implement diagnostic variables for collision rates
- [x] Test them against reproduction in Python

**Original objectives:**
_This PR adds new diagnostic variables for the energy and momentum transfer channels due to all the collisions in `collisions`. This should finally allow us to perform energy and momentum balances on a species basis. The collision frequencies are also saved with the new code `K`, e.g. `Kdd+_coll`. The rates for the various atomic processes are also going to be saved in the same format.
The frictional heating energy sources are saved separately due to the fact that both collision species end up with an energy gain. This prevents us from working out the energy source of species 2 from the inverse of the energy source of species 1, which is important as only the species 1 sources are saved.

In addition, the collision frequencies will be visible to other components and allow user choice in which collisions are used in the calculation of the neutral diffusion of particles, momentum and heat.

Tasks:

- [ ] Implement diagnostic variables for collision rates
- [ ] Implement diagnostic variables for collision energy and momentum transfer rates
- [ ] Allow the above to be grouped by default (i.e. source on one species due to ALL collisions)
- [ ] Implement diagnostic variables for frictional heating due to collisions
- [ ] Implement diagnostic variables for atomic rate frequencies
- [ ] Pass rates into the state and allow the neutral model to pick them up
- [ ] Test
- [ ] Add documentation_

### DIFF
--- a/include/collisions.hxx
+++ b/include/collisions.hxx
@@ -41,6 +41,9 @@ struct Collisions : public Component {
 
   void transform(Options &state) override;
 
+  /// Add extra fields for output, or set attributes e.g docstrings
+  void outputVars(Options &state) override;
+
 private:
   BoutReal Tnorm; // Temperature normalisation [eV]
   BoutReal Nnorm; // Density normalisation [m^-3]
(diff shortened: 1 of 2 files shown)


mikekryjak:
This PR is now scaled back to just the collision frequencies due to a change in priorities. The other objectives will be picked up by a separate PR.
@bendudson good to go

bendudson (approved):

PR #174, opened 2023-08-22
Fix recycling model flags in the examples

mikekryjak:
- Missed target_recycle=true
- Also adding recycle energy


mikekryjak:
Same changes replicated in https://github.com/bendudson/hermes-3/pull/176 in the hope that the documentation build doesn't break.

PR #175, opened 2023-08-24
Allow neutral cooling due to wall reflection

mikekryjak:
When hitting a wall, a neutral atom can either undergo thermal reflection (recombine, thermalise with the wall and come off as a molecule) or fast reflection (the atom does not recombine and bounces off, possibly losing some energy in the process). In EIRENE the fate of the particle is determined from the TRIM database.

The only thing we have on this right now is the `neutral_boundary` component which is currently only implemented for the sheath. It uses a sheath heat transmission style expression to allow for neutral energy to be imparted on the wall given a `gamma` parameter. The velocity used for this is the thermal velocity.

This PR adds a diagnostic variable for the heat loss due to this process and extends it to the SOL and PFR edges.

The expression will be following that from D. Power's 2023 thesis, and which was based on Horsten's work.
A sheath-style expression with a thermal velocity in one direction ( sqrt(Tn/Mn) ).
The gamma is calculated as:

gamma_n = 1 - alpha*R_r - (1-R_r)(T_fc / 2*Tn)

Where:

- Alpha is the fraction of energy particles retain after reflection
- R_r is the fraction of particles undergoing fast reflection as opposed to thermal reflection
- T_fc is the Franck-Condon dissociation temperature (about 3eV)
- Tn is the neutral temperature

Solving the above with alpha = 0.75 and R_r 0.8 results in a gamma of 0.3 for Tn = T_fc.
The gamma increases to 0.4 when Tn >> T_fc. The alpha and R_r coefficients were obtained by D. Power for typical conditions in deuterium-tungsten reflection.

Currently gamma is an input, but it has a dependency on Tn and should therefore be a calculation.
The inputs will be alpha and R_r now, with defaults of 1 and 1 (fully reflective with no energy loss)

Tasks:
- [x] Add diagnostic variable for the reflection energy loss: Ed_wall_refl (opinions on name?)
- [x] Extend to SOL and PFR edges
- [x] Modify to have energy and particle reflection coefficient as inputs
- [x] Test
- [x] Add documentation
- [x] Final check: full scale model

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -787,8 +787,15 @@ is set on parallel velocity and momentum. It is a species-specific
 component and so goes in the list of components for the species
 that the boundary condition should be applied to.
 
-An energy sink is added to the flux of heat to the wall, with
-heat flux `q`:
+A source of neutral cooling is added in accordance with the approach in the thesis of D.Power 2023.
+The source represents two kinds of neutral reflection:
+
+- Fast reflection, where a neutral atom hits the wall and reflects having lost some energy,
+- Thermal reflection, where a neutral atom hits the wall, recombines into a molecule, and then
+  is assumed to immediately dissociate at the Franck Condon dissociation temperature of 3eV.
+
+The energy sink has a heat flux `q` calculated from thermal velocity at the wall and a 
+heat transmission coefficient:
 
 .. math::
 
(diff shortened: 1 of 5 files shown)


mikekryjak:
I've reviewed Dom's thesis a bit more and found that the expression for gamma is actually a function of neutral temperature. I will make modifications to the implementation - I put some details in the main post.

mikekryjak:
Found a mistake in the calculation of the radial area. The toroidal length had a simple error - I didn't derive the equation for the average of the final cell and guard cell values and followed the convention from the poloidal length, which does not apply. I made sure this is now correct by reconstructing the heat flux in Python for the SOL and PFR. I assumed the targets are correct from before.

I changed the inputs so that we set the energy reflection factor and the fast reflection fraction and the gamma is calculated from those and the neutral temperature. I also added documentation. 

I've got a full res case running with the changes. The PR will be ready to go once that's OK.

bendudson (approved):
Thanks @mikekryjak , amazing work!

PR #184, opened 2023-11-23
Add reference to xHermes at the top of readme

mikekryjak:
Add reference to xHermes at the top of readme


bendudson (approved):
Thanks @mikekryjak

PR #185, opened 2023-11-28
Multigroup neutral model

mikekryjak:
The neutral fluid approximation breaks down in the core where the neutrals are so hot they are nearly collisionless. 

We assume a single neutral distribution. This means that those hot neutrals have no choice but to conduct their heat (preheat) to the nearby cold SOL neutrals, even though in reality they would pass through and hit the wall. To mitigate this issue, we can impose a conduction heat flux limiter, but this traps the hot neutrals in the core and prevents them from heating the walls, while the cold neutral preheat is okay in the SOL but too low in the PFR.

A solution could be to have a new species called hot neutrals. Both cold and hot neutrals would have the same atomic reactions available, but cold neutrals are created only through puffing and thermal recycling, while hot neutrals are created only through charge exchange or fast recycling.

To do:

- [x] Implement ionisation for hot neutrals
- [x] Implement transfer CX reaction
- [x] Simple test (presence of d*, new reactions: hot CX, transfer CX, hot IZ)
- [x] Implement neutral reflection d->d* transfer channel
- [ ] Test
- [ ] Extend to mixed isotope reactions
- [ ] 2D test with puff and pump
- [ ] Documentation


mikekryjak:
This PR will be abandoned. I got it to the point where it ran and the hot and cold species made sense, but I couldn't get a result that was more accurate than a single fluid in the limited time I had to work on this.

PR #187, opened 2023-11-29
Add tokamak_core component for user-friendly core boundary conditions

mikekryjak:
A common 2D/3D modelling workflow includes setting power and particle sources in the core. Hermes-3 currently allows:

- The core set through a constant temperature/pressure/density
- Core sources imposed by making appropriate Field2D arrays of sources on a per-cell basis and putting them in the grid file

The former is not very useful if you are matching a condition with known power and particle input, which is common. The latter is not very practical because to change the sources you need to modify the grid every time.
This PR adds a component called `tokamak_core` which aims to resolve the above with the following features:

- Ability to input core power and particle sources in W and s^-1 in the input file 
- Ability to make the core density and pressure boundary `reflective`, `escape` or `constant_value` 
- _Future work: enable ionisation and return of any escaping neutrals as ions_

The above is done on a per-species basis and will do away with the need to specify `Dirichlet` or `Neumann` manually. The intent is to implement more of these "physical" boundary conditions for the SOL and the target in the future.

To do:

- [x] Make the new component and implement source functionality with basic diagnostics
- [ ] Test
- [ ] Implement the `reflective`, `escape` and `constant_value` options for the boundaries
- [ ] Test
- [ ] Write documentation


mikekryjak:
Bumping this PR as something to finish if I get some time for a QoL focus week later this year.

PR #190, opened 2023-12-02
WIP: Boundaries and misc fixes

bendudson:
Investigating an anomaly in the temperature near the sheath: Sometimes the temperature ticks up in the last few grid cells.


mikekryjak:
@bendudson does this fix the temperature uptick? Was it the force balance? I see that it was setting the boundary erroneously from your commits, but I can't quite picture the physical consequence of that.

PR #191, opened 2023-12-06
Stability improvements and bug fix in neutral wall cooling formulation 

mikekryjak:
**Fixing cause of crash**
In testing the multigroup neutral model I found that I couldn't start my 1D baseline test case with both neutral momentum and the neutral boundary enabled. The neutral boundary contains fast/thermal reflection. The symptom was a `-4` crash at `t=0`, which I've also seen before in 2D and elsewhere.

I dug into it a bit and found that the cooling source q was a `nan:`
https://github.com/bendudson/hermes-3/blob/ea5c52d04206424cc18ed8b6d70a1734f7af4064/src/neutral_boundary.cxx#L109

This was because of two instances of division by zero when  `Tn=0` during the first step:
https://github.com/bendudson/hermes-3/blob/ea5c52d04206424cc18ed8b6d70a1734f7af4064/src/neutral_boundary.cxx#L91

https://github.com/bendudson/hermes-3/blob/ea5c52d04206424cc18ed8b6d70a1734f7af4064/src/neutral_boundary.cxx#L105-L106

Apart from flooring which probably should be a last resort option, the former can be eliminated by changing to a Neumann boundary (this would hopefully make a minimal difference to the solution) and the latter by rewriting `q` in terms of the particle flux which causes `Tn` to drop out. Technically one could avoid this by a better initial condition, but it is possible to have a zero temperature/density/pressure through solver error anyway.

**Changes to formulation**

There were two mistakes in the equation, see below. Even though a factor of 2 sounds scary, the net effect is much smaller, on the order of 10% additional wall cooling heat flux. However, neutral temperature can drop by more than that near the walls.

image

image

image

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -890,27 +890,40 @@ is set on parallel velocity and momentum. It is a species-specific
 component and so goes in the list of components for the species
 that the boundary condition should be applied to.
 
-A source of neutral cooling is added in accordance with the approach in the thesis of D.Power 2023.
-The source represents two kinds of neutral reflection:
+Just like ions can undergo fast and thermal recycling, neutrals can undergo fast or thermal 
+reflection at the wall. In edge codes using the kinetic neutral code EIRENE, this is typically
+controlled by the `TRIM database <https://www.eirene.de/old_eirene/html/surface_data.html>`_.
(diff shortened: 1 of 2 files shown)


mikekryjak:
@bendudson ready for review

bendudson (approved):

PR #192, opened 2023-12-09
Allow solving in 2D/3D without neutral momentum equation

mikekryjak:
In 1D neutral momentum is controlled by the `evolve_momentum` which can be easily disabled. In 2D, it is controlled by `neutral_mixed` which has the entire neutral transport model inside it and no option to disable it. This PR adds a new flag called `evolve_momentum` which is `True` by default. 

- [x] Implement flag
- [x] Test


mikekryjak:
@bendudson ready for review

PR #195, opened 2023-12-13
Implement collisionality selection to allow pure Braginskii closure

mikekryjak:
**Intro**
Closure refers to, among other things, the formulations for particle diffusion, conduction and viscosity. Standard Braginskii closure is only valid when there is only one heavy ion in the system (i.e. no impurities or other species like He and T). The current gold standard is Zhdanov closure (see Makarov 2021, 2023) which is implemented in all the other leading edge codes. Zhdanov is complicated and we don't have it yet.

In order to do a bit better than Braginskii, Hermes-3 currently keeps the Braginskii formulations but considers all enabled collisions (en, nn, ni) and charge exchange for a particular species. This at least enables some accounting of secondary ion species. However, it is likely really no substitute for Zhdanov and we must implement it as soon as we can.

While the current closure setup is better than Braginskii in a multispecies scenario, it makes code comparison and verification tricky. This PR implements the ability to switch to pure Braginskii for this reason. For neutral diffusion, the SOLPS AFN (see Horsten 2017) model considers CX and IZ rates and this is also reproduced here. 
Note: To preserve reasonable collisionality in a multi-ion setup, ei and ii collisions as well as charge exchange consider all available ions.

Here are the Braginskii collision choices for each process:

- Neutral diffusion: CX, IZ
- Electron conduction: ee
- Ion conduction: ii
- Ion viscosity: ii

**This PR:**

1. Adds diagnostics for all collision frequencies, e.g. `Kdd+_coll` is the ni collision frequency, `Kdd+_cx` is the d+ CX collision frequency and `Kd+_cond` is the overall collision frequency used for ion conduction. These are also now saved in the state under e.g. `species["collision_frequencies"]["d_d+_iz"]`.
2. Every time a bit of Braginskii closure needs to use a collision frequency, an extremely clunky string parser decides which ones to use. `multispecies` is the legacy behaviour where all collisions involving that species is included. `braginskii` is the classical Braginskii behaviour where conduction and viscosity use self-collisions only (e.g. ii for ions). For neutrals, the choice is `multispecies` or `afn`, where `afn` refers to the AFN (Advanced Fluid Neutral) model from Horsten 2017 and includes only CX and IZ collisions.
3. The chosen collision frequencies are printed to the console dump when the simulation starts. 

Originally, this PR had several other changes:
- Fix for a sqrt(2) mistake in electron collision time and the incorrect kappa constant for neutrals: now merged in https://github.com/boutproject/hermes-3/pull/302
- Refactor of `sheath_boundary_simple` for clarity/tractability and extra diagnostics: now merged in https://github.com/boutproject/hermes-3/pull/326

The PR features no tests because of time constraints, but was validated against ReMKiT1D as well as SOLPS/SOLEDGE2D in two of the ongoing benchmarks. The comparison to ReMKiT1D will become a regression test at some point. The current performance can become a regression test as well.

What's missing and left to further work:

- Tests (this will be done as part of the upcoming closure refactor)
- Implementation of collisions for impurity transport
- Refactoring of the string parser - see note below.

**Note: the collision selector has been done in a really awful way - it needs refactoring. Unfortunately I have no time to do this right now and even though it's ugly it works, so let's merge it and raise an issue. Ideally the collision strings would be prepared in one place not all over the code.**

To do:
- [x] Allow `collisions.cxx` to save individual frequencies
- [x] Allow hydrogenic ionisation and CX to save individual frequencies
- ~[ ] Implement for Ne and ADAS reactions?~
- [x] Allow conduction and viscosity to pick individual frequencies
- [x] Allow the neutral transport model to pick individual frequencies in 1D
- [x] Allow the neutral transport model to pick individual frequencies in 2D
- [x] Make sure this handles lack of collisions with an Exception
- [x] Test
- [x] Add to `evolve_energy.cxx`
- [x] Test with evolve_energy
- [x] Add documentation
- [x] Add documentation on tau_ee choice and references
- [x] Rename "legacy" to "multispecies", also in docs
- [x] Merge master into this PR
- [x] Separate neutral and ion parallel conduction collisionalities in evolve_pressure
- [x] Separate neutral and ion parallel conduction collisionalities in evolve_energy
- [x] Test on evolve_energy
- [x] Check ion_viscosity in multispecies mode


mikekryjak:
To check: ensure there are any instances of recombination/ionisation adding collision frequencies towards the electrons

mikekryjak:
Currently if no collisions are found, there will be a console print going forever like this:

> e conduction collisionality mode: 'braginskii' using 

The code needs to handle the lack of collisions with an exception when Braginskii is chosen.

mikekryjak:
Need to decide whether Braginskii or legacy should be default. Arguably fewer people do multi-ion stuff, so maybe Braginskii should be the default setting?

bendudson:
I think we should implement the collisions so that the rates are Braginskii when only one ion species is used. Charged particle collisions in the `collisions` component uses the Hinton formula (https://hermes3.readthedocs.io/en/latest/components.html#collisions) that is the same as Fitzpatrick's Braginskii collision frequency for e-i and i-i collisions.

mikekryjak:
Agreed. Implementing that on top of the way I've currently implemented the collision selector may get even more messy though. Could you take a look at what I've done and see how it can be simplified? My C++ is still not great yet. I am wary of merging this PR because of how janky the system is. Maybe the collision selector class needs to perform all of the selection within it, and `evolve_pressure` could just ask for a conduction rate, reducing its footprint there to one line.

The issue is that I don't have a lot of time for refactoring at the moment - I am behind on a lot of stuff.

mikekryjak:
@bendudson ready for review/merge

mikekryjak:
This isn't trivial to merge - it's been merged with `sheath-diagnostics` and had the `suppress_outflow` functionality implemented which have now been superseded by the implementation in `parallel-sheath`. I need to untangle that bit from it before we merge it in. 

Also note that this PR features a lot of line changes; I refactored the flux/flow calculations so that they are more clear (just like in `neutral_boundary`)

bendudson (approved):
Thanks, @mikekryjak !

PR #197, opened 2023-12-13
Allow for classical_diffusion to support user-set diffusion coefficient

mikekryjak:
Currently working with Chris Ridgers on simple Blob2D example. We need to add some kind of diffusion to it, but we don't want to have to evolve collisions. In this PR I modify the `classical_diffusion` component and add a new setting for a user-set D which bypasses collisions:

https://github.com/bendudson/hermes-3/blob/5d1e3811a7eef4da0d394a37cae97d1de5e1797c/src/classical_diffusion.cxx#L12

This then feeds into particle, momentum and pressure diffusion, but there is an additional bit (possibly conduction?) which has a different formulation. I haven't been able to figure it out yet so I disabled it for now - will need to discuss with @bendudson.

https://github.com/bendudson/hermes-3/blob/5d1e3811a7eef4da0d394a37cae97d1de5e1797c/src/classical_diffusion.cxx#L82-L85

### DIFF
--- a/include/classical_diffusion.hxx
+++ b/include/classical_diffusion.hxx
@@ -15,6 +15,7 @@ private:
 
   bool diagnose; ///< Output additional diagnostics?
   Field3D Dn; ///< Particle diffusion coefficient
+  BoutReal custom_D; ///< User-set particle diffusion coefficient override
 };
 
 namespace {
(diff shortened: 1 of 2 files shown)


mikekryjak:
The diffusion works, just need to figure out that last pressure equation term.

mikekryjak:
After discussing with Ben:
- Leftover energy equation term is conduction
- Leave it alone and document that the custom_D is for particle diffusion only
- In the future we can implement a custom kappa input if needed

PR #199, opened 2024-01-08
Draft: PI controller for upstream and/or target temperatures

tbody-cfs:
Figure plotted from example case
control

> I'd like to implement something similar to the density_controller for the electron and ion temperatures. I've started by copy-pasting the density_controller and trying to get the code to compile — there's a few mystery errors that I'm trying to work out.
> 
> Outstanding things to fix
> 
> 1. Need to make sure units make sense and are being used correctly. From `fixed_temperature` there's `SI::q_e` being used to convert between J and eV. However, including this term gives a compiler error that the `SI` library isn't found.
> 2. Are we correctly modifying and diagnosing the pressure source?
> 3. Add to documentation
> 4. Add tests


mikekryjak:
Hi @tbody-cfs! It would be awesome if it could also control to a set target temperature. That would allow the user to automatically obtain the detachment threshold condition.

One of the errors showing up in the CI test seems to be to do with SI constants. To make those work you need to have the below on the top of the file:

`#include <bout/constants.hxx>`

tbody-cfs:
How to fix this error?
```
In file included from /Users/tbody/Code/cfs-bout/hermes-3/hermes-3.cxx:74:
/Users/tbody/Code/cfs-bout/hermes-3/include/upstream_temperature_feedback.hxx:170:48: error: conflicting declaration 'RegisterComponent<UpstreamTemperatureFeedback> {anonymous}::register_uds'
  170 | RegisterComponent<UpstreamTemperatureFeedback> register_uds("upstream_temperature_feedback");
      |                                                ^~~~~~~~~~~~
In file included from /Users/tbody/Code/cfs-bout/hermes-3/hermes-3.cxx:73:
/Users/tbody/Code/cfs-bout/hermes-3/include/upstream_density_feedback.hxx:167:44: note: previous declaration as 'RegisterComponent<UpstreamDensityFeedback> {anonymous}::register_uds'
  167 | RegisterComponent<UpstreamDensityFeedback> register_uds("upstream_density_feedback");
      |                                            ^~~~~~~~~~~~
```

include/upstream_temperature_feedback.hxx
```diff
@@ -0,0 +1,173 @@
...
+
+  BoutReal source_multiplier; ///< Factor to multiply source
+
+  BoutReal proportional_term, integral_term; ///< Components of resulting source for diagnostics
+
+  bool diagnose; ///< Output diagnostic information?
+};
+
+namespace {
+RegisterComponent<UpstreamTemperatureFeedback> register_uds("upstream_temperature_feedback");
```
bendudson, line 170: The name "register_uds" conflicts with the definition in `upstream_density_feedback`, when the headers are included together. Just change to "register_uts" or other unique name.

tbody-cfs: Thanks! It's compiling now

include/upstream_temperature_feedback.hxx
```diff
@@ -0,0 +1,178 @@
...
+    // Source shape the same as used in EvolvePressure
+    pressure_source_shape =
+      alloptions[std::string("P") + name]["source_shape"]
+            .doc("Source term in ddt(P" + name + std::string("). Units [Pa/s], note P = 2/3 E."))
+            .withDefault(Field3D(0.0))
+        / (SPnorm);
+
+    diagnose = options["diagnose"]
+                   .doc("Output additional diagnostics?")
+                   .withDefault<bool>(false);
```
tbody-cfs, line 66: Something that I can't work out how to fix: with these changes the temperatures are no longer written into the output file by default. I need to add `diagnose = true` to each of the species (`e`, `d`, `d+`). This is the case even if I'm not using the temperature controller. I could change this line to `true` which would presumably fix the problem, but I'd prefer to understand why it's occuring.
@bendudson @mikekryjak thoughts?

mikekryjak: The `diagnose` flag has been set up on a per-species basis throughout the code, so this is expected behaviour. The `options` on line 64 is the options for the species and picks up settings immediately below the species header, e.g. `[d+]`.

src/upstream_temperature_feedback.cxx
```diff
@@ -0,0 +1,60 @@
...
+    source_multiplier = proportional_term + integral_term;
+
+    if ((source_multiplier < 0.0) && temperature_source_positive) {
+      source_multiplier = 0.0; // Don't remove particles
+    }
+
+    temperature_error_last = error;
+    temperature_error_lasttime = time;
+    break;
+  }
```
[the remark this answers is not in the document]
bendudson: Good question. This might be left in because in a 1D domain there's only one point at the boundary. The loop is iterating in X (radial coordinate) over all point at the lower Y boundary. In 1D simulations only the processor at the lower Y (upstream) end of the domain will enter the loop. It should exit after one iteration, because there's only one point on the boundary, but this `break` is probably there to make sure.

In 2D or 3D simulations we'd have to consider which temperature or density should be taken. Probably the maximum? But in that case the Lower Y boundary would be one of the targets. Probably need some more general strategy if we want to control input power in 2D simulations.

tbody-cfs:
# Modified 1D recycling example

Seems to be working. Temperatures pin to their set values, even as the density source varies


image

```
# 1D system with:
#  - no-flow boundary on lower Y
#  - sheath boundary on upper Y
#  - Evolving electron and ion species
#  - heat conduction
#  - Uniform source of heat and particles throughout domain
#  - Non-uniform grid, packed towards the target
#  - Recycling of ions as atoms
#  - Ionisation of neutrals as ions
#  - Charge exchange between neutrals and ions
#  - Feedback control of upstream density
#
#  Does not include recombination


nout = 1000
timestep = 200

MXG = 0  # No guard cells in X

[mesh]
nx = 1
ny = 400   # Resolution along field-line
nz = 1

length = 30           # Length of the domain in meters
length_xpt = 10   # Length from midplane to X-point [m]

dymin = 0.1  # Minimum grid spacing near target, as fraction of average. Must be > 0 and < 1

# Parallel grid spacing
dy = (length / ny) * (1 + (1-dymin)*(1-y/pi))

# Calculate where the source ends in grid index
source = length_xpt / length
y_xpt = pi * ( 2 - dymin - sqrt( (2-dymin)^2 - 4*(1-dymin)*source ) ) / (1 - dymin)

ixseps1 = -1
ixseps2 = -1

[hermes]
# Evolve ion density, ion and electron pressure, then calculate force on ions due
# to electron pressure by using electron force balance.
components = (d+, d, e,
              sheath_boundary, collisions, recycling, reactions,
              electron_force_balance, neutral_parallel_diffusion)

loadmetric = false        # Use Rxy, Bpxy etc?
normalise_metric = true  # Normalise the input metric?

Nnorm = 1e20
Bnorm = 1
Tnorm = 100

[solver]
type = cvode  # Backward Euler steady-state solver
use_precon = true

mxstep = 5000

atol = 1e-7
rtol = 1e-5

[sheath_boundary]

lower_y = false
upper_y = true

[neutral_parallel_diffusion]

dneut = 10   # (B / Bpol)^2 in neutral diffusion terms

####################################

[d+]  # Deuterium ions
type = (evolve_density, evolve_pressure, evolve_momentum,
        noflow_boundary, upstream_density_feedback, upstream_temperature_feedback)

noflow_lower_y = true
noflow_upper_y = false  # Sheath boundary at upper y

charge = 1
AA = 2

density_upstream = 1e20  # Upstream density [m^-3]
density_source_positive = false  # Force source to be > 0?
density_controller_i = 1e-1
density_controller_p = 5000.0

temperature_upstream = 110.0  # Upstream temperature
temperature_source_positive = false  # Force source to be > 0?
temperature_controller_i = 1e-2
temperature_controller_p = 10.0

thermal_conduction = true  # in evolve_pressure

diagnose = true

recycle_as = d
target_recycle = true
target_recycle_energy = 3.5  # Franck-Condon dissociation energy
target_recycle_multiplier = 1  # Recycling fraction

[Nd+]

function = 1

source_shape = H(mesh:y_xpt - y) * 1e20 # Feedback controller source shape

[Pd+]
function = 1

source_shape = H(mesh:y_xpt - y) * 1e9 # Feedback controller source shape

[NVd+]

function = 0

####################################

[d]  # Deuterium atoms
type = (evolve_density, evolve_pressure, evolve_momentum,
        noflow_boundary)

charge = 0
AA = 2

diagnose = true

thermal_conduction = true

[Nd]

function = 0.001

[Pd]

function = 0.0001

####################################

[e] # Electrons
type = quasineutral, evolve_pressure, zero_current, noflow_boundary, upstream_temperature_feedback

noflow_upper_y = false

charge = -1
AA = 1/1836

temperature_upstream = 100.0  # Upstream temperature
temperature_source_positive = false  # Force source to be > 0?
temperature_controller_i = 1e-2
temperature_controller_p = 10.0

diagnose = true

thermal_conduction = true  # in evolve_pressure

[Pe]

function = `Pd+:function`  # Same as ion pressure initially

source_shape = H(mesh:y_xpt - y) * 1e9 # Feedback controller source shape

####################################

[recycling]

species = d+

[reactions]
type = (
        d + e -> d+ + 2e,     # Deuterium ionisation
        d+ + e -> d,          # Deuterium recombination
        d + d+ -> d+ + d,     # Charge exchange
       )

```

src/upstream_temperature_feedback.cxx
```diff
@@ -0,0 +1,60 @@
...
+
+    if ((temperature_error_integral < 0.0) && temperature_integral_positive) {
+      // Limit temperature_error_integral to be >= 0
+      temperature_error_integral = 0.0;
+    }
+
+    // Calculate source from combination of error and integral
+    integral_term = temperature_controller_i * temperature_error_integral;
+    proportional_term = temperature_controller_p * error;
+    source_multiplier = proportional_term + integral_term;
```
tbody-cfs, line 39: Are we dropping the units of the temperature and time here? `integral_term` and `proptional_term` are meant to be dimensionless. Does that imply that `temperature_controller_i` has units of [time]^-1 * [temperature]^-1 and `temperature_controller_p` has units of [temperature]^-1?

tbody-cfs:
Something that doesn't make a lot of sense. The `"SPe"` term doesn't also look like the `"SPe_feedback"` term. 
Comparing these two (blue is feedback, orange is total source).
This case worked (was controlling for upstream temperature)
image
This case didn't work (was controlling for target temperature)
image

I can't really tell what the difference is between the two, and not sure where to start debugging.

src/upstream_temperature_feedback.cxx
mikekryjak: The P and I components are quantities calculated from dimensional data, but their own dimensionality doesn't matter because they're just control signals. I don't think there's any benefit in normalising them unless I'm missing something.

mikekryjak: On your other comments:
**Add in the "d" of a PID (example code below in Python)?**
I've raised this with Ben before as it seemed like a no brainer. There is a fundamental issue in calculating the D parameter, but I can't remember what it is. @bendudson?

After working with the density controller and tuning it on several occasions the P and I parameters seem more than sufficient. In fact you can get perfect performance with just the P parameter in some cases as well. I think this is because you are controlling a source which is in the same location as your control variable (i.e. the source and the control are both upstream) which makes the system response very fast and makes the control very easy.

**Is it possible to set Pe = Pd+, for Pd+ controlled via feedback or vice-versa? Probably want to have a Qi/Qe ratio ratio than a Ti/Te ratio**
Good question. I think this would need to be implemented in the code. One suggestion would be to add an option for a coupled species and a heat source ratio. You could add the controller for `d+` only, then set `e` as the coupled species and give it a ratio of 1. The controller then could pick this up from the input file and create a corresponding heat source for `e` calculated from the `d+` heat source times 1.

mikekryjak:
`SPe` is an aggregate of all the sources of `Pe`, while `SPe_feedback` is only from the feedback controller. Would be good to check other sources. If you are only giving heat to `Pe`, then I would guess that there will be a lot of heat being lost to the ions. Unfortunately `Qei` is not yet implemented as a diagnostic variable.

On the case controlling to target temperature, have you put the code for that on a commit yet? I had a quick look and couldn't immediately see it.

tbody-cfs:
> `SPe` is an aggregate of all the sources of `Pe`, while `SPe_feedback` is only from the feedback controller. Would be good to check other sources. If you are only giving heat to `Pe`, then I would guess that there will be a lot of heat being lost to the ions. Unfortunately `Qei` is not yet implemented as a diagnostic variable.
> 
> On the case controlling to target temperature, have you put the code for that on a commit yet? I had a quick look and couldn't immediately see it.

Something very confusing, figured it out now. Setting a pressure source has no effect on the pressure source... we need to set an energy source.

tbody-cfs:
> Note to self: this will fail if controlling for the target temperature in parallel.
> 
> ```
> // The upstream temperature is only present in processor 0, so for other
>   // processors the above loop is skipped. Need to broadcast the values from
>   // processor 0 to the other processors
>   MPI_Bcast(&source_multiplier, 1, MPI_DOUBLE, 0, BoutComm::get());
>   MPI_Bcast(&proportional_term, 1, MPI_DOUBLE, 0, BoutComm::get());
>   MPI_Bcast(&integral_term, 1, MPI_DOUBLE, 0, BoutComm::get());
>   ASSERT2(std::isfinite(source_multiplier));
> ```

Last processor should be `BoutComm::size()-1`, not `BoutComm::size()` right?

bendudson (approved):
Awesome work, thanks @tbody-cfs !

PR #200, opened 2024-01-12
Add simplified Ar cooling curves

mikekryjak:
Impurity radiation dominates detachment in reactor-class devices but it's a very nonlinear process partially because of the nonlinearities of the cooling curve:
- Both the LHS and RHS slopes are curved 
- RHS slope tapers off slower than the LHS slope
- RHS slope has a large "shoulder", a region of non-zero radiation at higher temperatures which is also a complex curve

All three features are present in curves of N, Ne, Ar and C to varying degrees. To study the impact of this, I prepared two simplified curves for Argon. `fixed_fraction_argon_simplified1` takes out the curvature from the slopes but preserves their levels:
image

`fixed_fraction_argon_simplified2` takes out the RHS shoulder:
image

`fixed_fraction_argon_simplified3` takes out the LHS - RHS asymmetry:
image

Note that there is an additional point in the LHS limit to smooth out the transition to no radiation in the low temperature limit. Sharp transitions there have led to long runtimes in the past.

- [x] Add the first two simplified curves
- [x] Test
- [x] Add the third curve
- [x] Test
- [x] Write docs

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -1834,6 +1834,11 @@ collisional-radiative model has been set to :math:`1\times 10^{20} \times 0.5ms`
 Each rate has an upper and lower bound beyond which the rate remains constant. 
 Please refer to the source code in `fixed_fraction_radiation.hxx` for the coefficients and bounds used for each rate.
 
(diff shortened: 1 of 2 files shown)


mikekryjak:
@bendudson ready for review

bendudson (approved):

PR #201, opened 2024-01-17
Add option to prevent neutrals receiving CX momentum

mikekryjak:
Adding an option to prevent ion momentum from being transferred to the neutrals during a CX reaction. The ions still lose momentum, but it simply doesn't go anywhere. This is normal behaviour when the neutral momentum equation is disabled (since we are not solving for the momentum it cannot be gained, leading to ion momentum being lost from the system).

When you enable the neutral momentum equation, there are two significant effects:
- Neutral transport is much more accurate because parallel kinetic energy is evolved and advected
- Charge exchange results in true momentum transfer from ions to neutrals instead of momentum destruction

In effect, this PR allows you to study the effect of just the former by disabling the latter. This can be done by setting `no_neutral_cx_mom_gain = true` under `[reactions]`.

To do 

- [x] Implement
- [x] test
- [x] docs


mikekryjak:
@bendudson this one's ready for review and merge. It's already been tested extensively as part of our STEP work.

bendudson (approved):

PR #204, opened 2024-02-01
Add simple 1D reservoir model

mikekryjak:
A major limitation of 1D SOL models is the inability to consider cross-field transport. This is particularly impactful in deep detachment where the neutral pressure can be many times higher than in a comparable 2D model where neutrals can spread across field lines and get pumped out.

This PR adds a simple reservoir model based on the one in DIV1D but a step simpler. The user defines the reservoir location by passing an expression to `reservoir_location` and sets it to a constant density with `reservoir_density`. The exchange term for density will be proportional to the difference between the domain density and the reservoir density, and there will be proportional pressure and momentum sinks. In regions where the particle transfer goes in the other direction, new particles will appear with the same pressure and momentum as the local ones.

Still trying to work out exactly what DIV1D did but I think they evolve the reservoir density with a domain exchange term as described above plus a pump sink term. This is according to Gijs Derks' MSc thesis and may have changed since then.

The reservoir can be implemented for any species.

- [x] Implement component
- [x] Implement exchange channels
- [ ] Think about whether we need to evolve density
- [ ] Think about what pressure and momentum the returning particles should be
- [ ] Fix description of energy source (currently pressure)
- [ ] Test
- [ ] Docs

If we have evolving reservoirs we need to evolve a scalar. Peter prepared an example: https://github.com/boutproject/BOUT-dev/blob/evolve-scalars/tests/integrated/test-solve-scalar/test_solve_scalar.cxx


src/reservoir.cxx
```diff
@@ -0,0 +1,220 @@
...
+      options["diagnose"].doc("Save additional diagnostics?").withDefault<bool>(false);
+          
+}
+
+
+void Reservoir::transform(Options& state) {
+  AUTO_TRACE();
+
+  // We are operating on only one species
+  auto species = state["species"][name];
```
bendudson, line 57:
```suggestion
  auto& species = state["species"][name];
```

src/reservoir.cxx
```diff
@@ -0,0 +1,220 @@
...
+  Field3D N = get<Field3D>(species["density"]);
+  Field3D P = get<Field3D>(species["pressure"]);
+  Field3D NV = get<Field3D>(species["momentum"]);
+
+
+  for(int ix=0; ix < mesh->LocalNx ; ix++){
+      for(int iy=0; iy < mesh->LocalNy ; iy++){
+          for(int iz=0; iz < mesh->LocalNz; iz++){
+
+            auto i = indexAt(N, ix, iy, iz);
```
bendudson, line 95: I think this triple-nested loop and `indexAt` can be replaced by:
```
BOUT_FOR(i, N.getRegion("RGN_NOBNDRY")) {
   ...
}
```

src/reservoir.cxx
bendudson: @ZedThree question: Is there a way that we can force copies to be explicit? Perhaps delete the copy constructor and add a `copy` function so that 
```
auto species = state["species"][name];
```
would be a compile error. If someone really wants to copy they have to make it explicit:
```
auto species = copy(state["species"][name]);
```
Having the behavior depend on a single `&` in the right place is a footgun that keeps going off.

mikekryjak:
There is an issue where the final domain cell is seeing an extremely high density source in the last domain cell. Here is a plot near the target which includes guard cells showing the model output and a post-processing calculation which is identical to what's in the code.

image

There is no large energy source there which is bizarre since it should be proportional. However, there is a large pressure source in the first guard cell for whatever reason.
image

The momentum source matches the calculation, unlike energy and density. Again, this is very strange since the momentum source is proportional to the density source:
image

No clue what's going on here. I think I will start by setting the guard cells to 0 in case they are having an impact, but I thought they wouldn't. @bendudson any ideas?

mikekryjak:
@PoloidalLloyd is this PR superseded by yours? Can we close it?

PoloidalLloyd:
Hi @mikekryjak , I have made some alterations to the implementation, although the base approach remains similar to DIV1D. To achieve this however it required coupling with @tbody-cfs's [field line geometry component](https://github.com/boutproject/hermes-3/pull/232) to function properly. I'll put together a PR this week, however not sure whether to include field line geometry component in this, it should probably be separate.

mikekryjak:
@PoloidalLloyd thanks for the update. https://github.com/boutproject/hermes-3/pull/232 is very close to being merged now, so as long as you merge that branch into yours there won't be a problem. Feel free to raise a PR with both if it makes it easier.

PR #207, opened 2024-02-21
Additional 2D neutral model flags for diagnostics

mikekryjak:
Added flag to disable neutral conduction and to make the perpendicular transport operators upwind


mikekryjak:
This PR will likely not be needed when https://github.com/bendudson/hermes-3/pull/212 is complete

Issue #196, opened 2023-12-13
Impurity transport: AMJUEL He and ADAS C, Ne reactions don't register collision frequencies


mikekryjak:
A species' collision frequency is used to calculate its conductivity, viscosity and (if it's a neutral) diffusive transport. These come from `collisions.cxx` and from atomic reactions of ionisation (IZ) and charge exchange (CX). In hydrogenic plasmas CX and IZ dominate neutral collisionality and are required for neutral transport to work well.

It looks like He as well as the ADAS reactions for C and Ne don't register their IZ or CX collision frequencies. For example, I would expect ADAS charge exchange to save its collision frequency in this block:
https://github.com/bendudson/hermes-3/blob/88f7748cfec058d71e386064f94ce5955369af2d/src/adas_reaction.cxx#L212-L229

Is this on purpose because most of these impurity species are ionised and their collisionality is dominated by `ei` or `ii` collisions? 
I am concerned that while this could be fine for the ions, the neutral species could be seeing extremely low collision frequencies, which would lead to an over-prediction of diffusive neutral transport. This would be particularly impactful in 1D where the diffusive transport is still without a flux limiter. 

I wonder whether resolving this could alleviate the performance issues when running impurity transport.
I am currently working on PR https://github.com/bendudson/hermes-3/pull/195 reworking the way the frequencies are saved into the state, so it's an opportunity to get this resolved following discussion @bendudson.

tbody-cfs:
Should look at this: can we set impurity neon CX similar to hydrogen CX?

PR #210, opened 2024-03-01
Add ADAS CX rate to species collision_frequency

bendudson:
The colliding species experience a collision that should impact e.g viscosity and heat conduction.

Note: Collisional processes are complex, and lumping everying together into a single collision frequency is a simplification.

Partly fixes issue #196

### DIFF
--- a/src/adas_reaction.cxx
+++ b/src/adas_reaction.cxx
@@ -247,4 +247,13 @@ void OpenADASChargeExchange::calculate_rates(Options& electron, Options& from_A,
     subtract(from_B["energy_source"], energy_exchange);
     add(to_B["energy_source"], energy_exchange);
   }
+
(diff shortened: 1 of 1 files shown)


mikekryjak:
@oparry-ukaea did your refactor already take care of this by any chance?

oparry-ukaea:
Hi @mikekryjak, 
Yes, switching the ADAS classes to inherit from Reaction should populate "collision_frequency" for all reactant species automatically.
RateHelper doesn't floor the density product / mass action factor at zero in the way that this change does though... does it need to? I thought evolve_density already floors all the densities to zero anyway?

mikekryjak:
Hmm, this is true, `evolve_density` floors densities to zero before saving them into the state. @bendudson how come we need to do this again elsewhere?

PR #212, opened 2024-03-05
neutral_mixed: Fix neutral floor scheme, make perp operator same as anomalous transport

bendudson:
~~Apply limiter/upwinding to both X and Y advection.~~

**MK edit**
In this PR:
- Add Div_a_Grad_perp with flow diagnostics and add them to the neutral model
- Fix consistency in flooring scheme in `neutral_mixed` and calculate `pn_floor` from `nn_floor` to prevent unphysical temperature spikes if pressure changes and density is limited by floor
- Add lax flux capability to parallel advection of density, momentum and pressure
- Allow disabling of neutral conduction
- Fix segfault if neutral momentum evolved but pressure not set, leading to Snv not being allocated here: https://github.com/bendudson/hermes-3/blob/88aa8c6c4d929643cfa9389b6a15ebb848d00634/src/neutral_mixed.cxx#L371-L374

Note that instead of providing total flows and warning that they may be incomplete, this implementation writes specific flows from specific processes. I think this is more useful because A) it provides certainty on what is and is not included in the diagnostic and B) provides more granular diagnostics which are useful for testing. I propose that the other components are updated accordingly over time. The granular diagnostics can always easily be merged into some sort of total using a flag to lighten the diagnostic data load.

Still to do:
- [x] Fix broken merge by adding parallel flows
- [x] Update flow diagnostic variables in line with new system
- [x] Add flow diagnostics to the other operators?
- [x] Improve formatting for legibility

Longer term issues:
- neutral transport is provided by evolve_density, evolve_pressure and evolve_momentum in 1D but by neutral_mixed in 2D. This should be resolved.
- The diagnostics are now complete for the energy and density equations, which should allow an easy particle/energy balance test based on the neutral_mixed integrated test.


src/neutral_mixed.cxx
```diff
@@ -363,9 +369,10 @@ void NeutralMixed::finally(const Options& state) {
       // eta_n = (2. / 5) * kappa_n;
       //
 
-      ddt(NVn) += AA * FV::Div_a_Grad_perp((2. / 5) * DnnNn, Vn)    // Perpendicular viscosity
-                + AA * FV::Div_par_K_Grad_par((2. / 5) * DnnNn, Vn) // Parallel viscosity
-        ;
+      ddt(NVn) +=
+          AA * FV::Div_a_Grad_perp((2. / 5) * DnnNn, Vn)      // Perpendicular viscosity
+          + AA * FV::Div_par_K_Grad_par((2. / 5) * DnnNn, Vn) // Parallel viscosity
```
mikekryjak, line 374: Hi @bendudson how come you haven't put limiters on perpendicular viscosity?

mikekryjak: Perpendicular conduction is also left out. It's true that we know that conduction helps and not hinders, but I'm tempted to change it all to upwind to see if it fixes the problem, and only then remove it from viscosity and conduction. What do you think?

mikekryjak: Also, I noticed that these operators are taking the same global compile time limiter as the other operators. I saw quite a significant impact on the results when I tried MinMod in a test case a few months back. It would probably be useful to have the ability to keep MC for most of the equations and have MinMod or Upwind or similar only for the neutral operators. 
But again, more than happy for everything to be on MinMod for the first test run if it fixes the problem..

bendudson: Hey @mikekryjak . I've started by only putting limiters on advection terms: The perpendicular pressure diffusion is an advection with a velocity proportional to the pressure gradient, not a diffusion. I've left the diffusion terms as center differences for now (viscosity and heat conduction).

bendudson: Changing to Upwind (from MC) seems to improve the result (and speed it up). Still failing the integrated test for some reason.

mikekryjak:
The integrated test is failing due to checkerboarding. See top for the master branch and bottom for this branch. This is odd because we switched to upwind operators...
image

mikekryjak:
@bendudson have you had a chance to look into this yet?

mikekryjak:
@bendudson we should try to merge this one in as this PR has the more robust 2D neutral transport flooring scheme.
We still need to switch back to the non-upwind operator and modify it to include flows, which will prevent checkerboarding and make it pass the test.

include/div_ops.hxx
```diff
@@ -23,12 +23,12 @@
 
  */
 
-#ifndef __DIV_OPS_H__
-#define __DIV_OPS_H__
+#ifndef HERMES_DIV_OPS_H
```
mikekryjak, line 26: Hi @bendudson I had a google about what this means and I see that it's to prevent the header file from being included more than once across the code. But how come you're changing the name here? This is a merge conflict and I'm not sure how to resolve it.

bendudson: It's just to prevent the file from being included twice. The name doesn't really matter, but names starting `__` are not supposed to be used, since they may conflict with compiler internals.

mikekryjak:
@bendudson I reverted the operator back to Div_a_Grad_par and added flows. The neutral_mixed integrated test passes on my end, but fails on GitHub with a code 8. I can't for the life of me find the exit code reference in CTest docs which are very brief.

mikekryjak:
Ready to review. 
Also, fixes https://github.com/bendudson/hermes-3/issues/267.

Issue #149, opened 2023-07-13
`cmake -E copy_directory` fails while editing


dschwoerer:
emacs uses symlinks to lock files that are currently edited.

`cmake -E copy_directory` chocks on those links, and the copy_directory fails without any explanation.

This should be avoidable by explicitly copying the files, rather than whole directories.

See also:
https://cmake.org/pipermail/cmake/2009-September/032219.html

bendudson:
Hi @dschwoerer Sorry I think the solution is either in CMake or emacs

PR #220, opened 2024-04-02
Avoid copy_directory

dschwoerer:
Use an explicit list, that is auto-generated

Resolves: #149


bendudson:
Thanks @dschwoerer for working on this. I'd prefer to keep the CMake commands as less code complexity to manage. I agree it's not ideal, but want to try and keep things simple if possible.

PR #223, opened 2024-04-30
Fix STPM component

### LINKED ISSUE #222: Prevent user from adding a pressure_source to state by mistake
We most commonly solve energy through a pressure equation, but we have only an `energy_source` and no `pressure_source`. While this makes sense if we want to be able to switch to a conservative energy equation, it is unintuitive and has led to several mistakes in the past. Can we raise an exception if `pressure_source` is set in the state?

@bshanahan @dschwoerer @bendudson

dschwoerer:
Also fixes #222

### DIFF
--- a/include/binormal_stpm.hxx
+++ b/include/binormal_stpm.hxx
@@ -31,12 +31,15 @@ struct BinormalSTPM : public Component {
   ///     - density correction
   ///
   void transform(Options& state) override;
+  void outputVars(Options &state) override;
 
 
 private:
   std::string name; ///< Short name of the species e.g. h+
+  bool diagnose; ///< Output diagnostics?
   Field3D Theta, chi, D, nu; ///< Field line pitch, anomalous thermal, momentum diffusion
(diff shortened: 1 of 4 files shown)


mikekryjak:
Thanks @dschwoerer. Could you give me a quick overview of how CHECKLEVEL works and why you decided to give it a level 1? What's the default and how do you change it (e.g. BOUT++ or Hermes-3 compilation flag?)
I've never worked with this before and it sounds like something I should really know about...

dschwoerer:
You can change the checklevel in BOUT++, by setting `-DCHECK=<0...4>` to 
cmake.
`-DCHECK=0` disabled mostly all checks, while `-DCHECK=4` makes the 
simulations much slower.
The default is `2` and should only have a few percent overhead.

I picked 1 because I think this check should by very fast. But it could 
probably also have been 2, as it is rather unlikely that this goes wrong.

bshanahan:
this fixes the issue. Thanks @dschwoerer @mikekryjak

bendudson (approved):
Thanks @dschwoerer !

PR #224, opened 2024-04-30
Fix radiation multiplier not affecting 13.6eV e heating in recombination

mikekryjak:
As per title. Ready for review


bendudson (approved):
Thanks @mikekryjak !

PR #232, opened 2024-06-19
Geometry 1D modifications

tbody-cfs:
*n.b. feel free to edit/comment if there are errors in the logic presented here*

# MK edit

This PR aims to significantly improve how we define the geometry of our 1D simulations.
It adds the following functionality:

- Ability to define input file parameters in terms of Lpar, including flux expansion
- Ability to define a profile of magnetic pitch
- Ability to define a profile of SOL broadening
- Calculation of flux tube radial cross-sectional area for communication with reservoirs
- Passing all of the above into the state for use in other components

# Accounting for cross-field transport and flux-expansion in 1D simulations

Simple models such as the extended Lengyel and DLS models show the importance of capturing cross-field transport (i.e. $\lambda_{INT}=\lambda_q + 1.64S$) and flux expansion (i.e. $B$). We want to be able to simply vary these parameters to study their effect and compare Hermes-3 to other models.

One implementation we might want to use as a reference is [Derks et al., 2024](https://doi.org/10.1088/1361-6587/ad2e37). In this paper, conservation of heat flux is used to give
$$\lambda(x) \cdot \sin(\theta)(x) \cdot B_{trans} = const.$$
where $\lambda$ is the poloidal width of the heat flux channel and $\sin(\theta)=B_{pol}/B_{tor}$. Let's introduce the shorthand $\sin(\theta)=\Theta=B_{pol}/B_{tor}$.

$B_{trans}$ is a modification to the actual magnetic field $B_{field}$ intended to mock up cross-field transport, such that the magnetic field in the model is set as $B_{model}=B_{field} \cdot B_{trans}$.

## Discussion with @bendudson 

In 1D, $J \cdot dx \cdot dy \cdot dz = V_{cell}$

where $dx = 1$ and $dz = 2 \pi$ by default.

These coordinates are set in https://github.com/boutproject/hermes-3/blob/master/hermes-3.cxx#L247.

Looking at this, we essentially can vary $J$ to set $A_\perp$.

$$J = 1/B_{model} = b / B_{actual}$$
would do what we're trying to achieve

## Final implementation

**Functionality of the `fieldline_geometry` Component**

The `fieldline_geometry` component in Hermes-3 is designed to handle the geometric calculations necessary for 1D simulations, particularly those aligned with magnetic field lines.

**Key Responsibilities:**

* **Parallel Length Calculation:** Computes the parallel distance (`lpar`) from an upstream reference point along the magnetic field line. This is a fundamental quantity for 1D simulations as it defines the spatial coordinate.
* **Geometric Quantities:** Calculates essential geometric quantities that vary along the field line:
    * Flux tube width.
    * Flux expansion.
    * Cell dimensions (poloidal length, side area, and volume).
* **Magnetic Field Handling:**
    * Accepts user-defined expressions for the poloidal (`Bpol`) and toroidal (`Btor`) components of the magnetic field.
    * Provides an option to compute the toroidal field component from the major radius (`R`), which is a common simplification in some contexts.
    * Calculates the total magnetic field strength (`B`) and the pitch angle (`\\(\\theta\\)`).
* **Jacobian Calculation:** Computes the Jacobian (`J`), which is crucial for coordinate transformations and volume integrals in the simulation.
* **Output:** Adds the calculated geometric quantities to the simulation state for use by other components and for diagnostic output.

**Equations:**

* Pitch angle:  $\sin(\theta) = B_{pol} / B$
* Toroidal magnetic field (if computed): $B_{tor} = B_{tor,up} R_{up} / R$
* Flux expansion: $f_{exp} = (B_{pol,up} / B_{up}) / (B_{pol} / B)$
* Flux tube width: $\lambda_{int} f_{exp}$
* Cell poloidal length: $dl_{pol} = dl_\parallel B_{pol} / B$
* Cell side area: $2 \pi R \cdot dl_{pol}$
* Cell volume: $2 \pi R \cdot dl_{pol} \lambda_{int} f_{exp}$
* Jacobian: $J = (\lambda_{int} / \lambda_{q,up}) / B$

**In essence:**

The `fieldline_geometry` component provides the geometric foundation upon which 1D field-aligned simulations in Hermes-3 are built. It handles the complexities of the magnetic geometry, allowing other components to focus on the physics of the plasma.


tbody-cfs:
# Test
broadening

[evidence.zip](https://github.com/user-attachments/files/19980119/evidence.zip)

tbody-cfs:
One thing I've found — if the magnetic field doesn't equal the jacobian, we end up with a non-1 Jacobian. This has a small but noticeable impact on the simulation. Should the Jacobian be normalized such that it is 1 if the magnetic field is constant?
image

tbody-cfs:
Full round of tests [test_fieldline_geometry.pdf](https://github.com/user-attachments/files/20015505/test_fieldline_geometry.pdf)

mikekryjak (changes requested):
Hi @**Tom Body**, really great work on this - very nice clear code with lots of comments, and I really appreciate the docs as well as the test pdf you sent. Once this is merged, I think we should calculate some similar variables for 2D as well, that's been on my list for ages.

My comments are mostly about diagnostic naming and documentation location. 

I think it would also be really good if we had a proper test as we discussed at the dev meeting last week. Do you think this is something you have time to do? If not, then could you make a test by slapping some asserts in a Python script which will read the non-trivial formulation case?

tbody-cfs:
Addressed @mikekryjak's review comments in e656907d:

- `pitch_angle` → `fieldline_geometry_magnetic_pitch`
- `fieldline_radius` → `fieldline_geometry_Rxy`
- `poloidal/toroidal/total_magnetic_field` → `fieldline_geometry_Bpxy` / `fieldline_geometry_Btxy` / `fieldline_geometry_Bxy`
- `flux_expansion` → `fieldline_geometry_f_R`
- `cell_poloidal_length` → `fieldline_geometry_dlpol`
- `cell_side_area` left as-is for now (no name agreed yet)
- Moved docs to immediately after `domain_grid` in the toctree

tbody-cfs:
Needs https://github.com/boutproject/hermes-3/pull/598 to pass clang-format test

mikekryjak:
Thanks @tbody-cfs, sorry I haven't had a chance to take a look yet - will do this week.

src/fieldline_geometry.cxx
```diff
@@ -0,0 +1,319 @@
...
+  lpar(0, 0, 0) = 0.5 * dy(0, 0, 0); // Initialize lpar at the first cell center
+  // The parallel length at the first cell center is half of the cell width.
+  for (int id = 0; id <= NYPE - 1;
+       ++id) { // Iterate through each processor in the poloidal direction
+    for (int j = 0; j <= mesh->LocalNy;
+         ++j) { // Iterate through each cell in the local poloidal domain
+      if (j > 0) {
+        // Calculate lpar at the current cell center by adding the average of the current
+        // and previous cell widths to the previous lpar value. This is a simple trapezoidal rule.
+        lpar(0, j, 0) = lpar(0, j - 1, 0) + 0.5 * dy(0, j - 1, 0) + 0.5 * dy(0, j, 0);
```
mikekryjak: Have you tested this when running in parallel? This looks reasonable on a single rank, but presumably each rank after the first one would need to accumulate on the connection length from the previous ranks. But I don't see you using `id` for anything in this loop.

src/fieldline_geometry.cxx
```diff
@@ -0,0 +1,319 @@
...
+  // Parallel length of cell
+  Field3D dlpar = coord->dy / Lnorm;
+  // Width of flux tube in the radial direction
+  flux_tube_width = lambda_int * flux_expansion;
+  // Length of the cell in the poloidal direction
+  cell_poloidal_length = dlpar * pitch_angle;
+  // Poloidal area of the cell (poloidal length times circumference)
+  cell_side_area = cell_poloidal_length * 2.0 * PI * fieldline_radius;
+  // Volume of the cell (poloidal area times flux tube width)
+  cell_volume = cell_side_area * flux_tube_width;
```
mikekryjak, line 246: I see that none of the new geometry variables are put into the state for other components to read. Is this intentional? An easy way to do this would be to add code in `transform_impl` to put them into the state at every RHS. A better way would be to do it in the `coordinates` object, but that would be a BOUT++ change and it would be more complicated. I personally think the former is sufficient.

I know this is already being called by the new reservoirs - how does that work? @PoloidalLloyd

src/fieldline_geometry.cxx
```diff
@@ -0,0 +1,319 @@
...
+  flux_expansion = pitch_angle(0, mesh->ystart, 0) / pitch_angle;
+
+  // Compute the effective magnetic field strength, which is the actual magnetic field strength divided by the transport broadening.
+  // This is used in the Jacobian calculation.
+  Field3D effective_magnetic_field_strength = total_magnetic_field / transport_broadening;
+
+  // Bxy is no longer consistent with the Jacobian.
+  // Set equal to NaN, to prevent anyone from using it.
+  BOUT_FOR(i, coord->Bxy.getRegion("RGN_ALL")) {
+    coord->Bxy[i] = std::numeric_limits<BoutReal>::quiet_NaN();
```
mikekryjak, line 218: Hmm... I think `Bxy` almost certainly isn't used for any physics relevant to 1D simulations, but not 100% sure. @bendudson what do you think?

mikekryjak (changes requested):
This is great work @tbody-cfs, especially the tests! I have a few questions, see below.

src/fieldline_geometry.cxx
tbody-cfs: Good catch — I tested this in the Docker image, running a non-trivial 1D case at 1, 4 and 8 ranks and comparing both the geometry outputs and the evolved solution cell-by-cell against the single-rank run. It surfaced two genuine parallel bugs.

On the loop itself / id: the accumulation is correct across ranks, but you're right that id wasn't doing anything. The mechanism is that each mesh->communicate(lpar) carries the running integral one rank downstream, so repeating the integrate-then-communicate step NYPE times propagates the value to every rank. id was just the repeat counter — I've renamed it sweep and documented that. I also fixed an off-by-one out-of-bounds write in the inner loop (j <= mesh->LocalNy → j < mesh->LocalNy). With this, lpar matches the single-rank result to machine precision.

Bug 1 — upstream reference used each rank's local first cell. Several quantities referenced field(0, mesh->ystart, 0), which is the local first cell, not the global upstream. So f_R, transport_broadening and Btor (with compute_Btor_from_R) — and hence the Jacobian — were only right on rank 0; every other rank restarted f_R at 1.0 at its local boundary:

f_R, np=4, first cells of each rank vs single-rank reference:
rank 0: [1.000 1.000 1.001 1.003]  ref [1.000 1.000 1.001 1.003]  ✓
rank 1: [1.000 0.997 0.994 0.991]  ref [1.849 1.843 1.837 1.830]  ✗
rank 2: [1.000 0.997 0.994 0.991]  ref [1.200 1.197 1.193 1.189]  ✗
rank 3: [1.000 1.000 0.999 0.998]  ref [1.016 1.016 1.015 1.015]  ✗

Fixed with an upstream_value() helper that reads the reference on rank 0 and MPI_Bcasts it. After this, all geometry outputs (lpar, Rxy, lambda_int, f_R, dlpol, Bxy) match the serial run to machine precision (max rel diff 0.0) at np = 1/4/8.

Bug 2 — coord->J was set only on interior cells. The geometry outputs matching wasn't sufficient: comparing the evolved solution showed a larger np1-vs-np4 spread for the default 1d-threshold case. The Jacobian was written only for ystart..yend, so the guard cells kept the default mesh J, discontinuous with the fieldline J. I now set J over the whole local field (guard cells included) and mesh->communicate(coord->J), mirroring recalculate_metric. Max relative np1-vs-np4 difference at a fixed output step:

field   before J-fix   after J-fix   1d-threshold
Ne         1.30%          0.57%          0.57%
Td         2.49%          1.49%          1.47%
Nd         1.60%          0.80%          0.80%

i.e. the fix brings the parallel spread down to the baseline the 1d-threshold already finds. Added a unit-test assertion that J is set in the guard cells (it fails against the old interior-only code).

On the residual difference: the ~0.5–1% that remains is not geometry-specific — the 1D-threshold shows the same, because this is the beuler steady-state solver at a fixed output step (snes_local_residual ~ 1e-3); the two decompositions take slightly different internal steps and land at slightly different points on a still-evolving transient. Driven to a tighter residual they converge to the same steady state.

src/fieldline_geometry.cxx
tbody-cfs: Not intentional — good suggestion, and I've taken the transform_impl approach you describe (agree the coordinates-object route is more invasive than it's worth here). transform_impl now publishes the geometry quantities into the shared state on every RHS, with matching readWrite permissions declared in the constructor, so other components can get<Field3D>() them. The keys mirror the diagnostic names — fieldline_geometry_Rxy, _Bxy, _lpar, _dlpol, _cell_volume, etc. Added a unit test (TransformPublishesGeometryToState) that checks all of them are published.

On the reservoirs — I don't have that component on this branch, so nothing consumes these yet here; this just makes them available. @PoloidalLloyd, could you confirm which keys the reservoir reads so I can make sure the names line up? Happy to rename to match whatever it expects.

src/fieldline_geometry.cxx
tbody-cfs: Looks like Bxy consumer is used in perpendicular-dynamics / turbulence / EM components (vorticity, the perpendicular operators in div_ops, diamagnetic & polarisation drifts, classical + ion/electron viscosity, neutral_full_velocity, electromagnetic, and the Alfvén-speed term in sound_speed)

tbody-cfs:
To test this out yourself @mikekryjak

# Testing `fieldline_geometry` in parallel with the Docker image

This walks through building the `geometry_1d` branch inside the
`ghcr.io/boutproject/hermes-3` image and verifying that a 1D
`fieldline_geometry` case gives the **same answer** regardless of the
y-decomposition (number of MPI ranks). Everything is copy-pasteable.

The idea: `nx=1` / `MXG=0` forces `NXPE=1`, so all MPI ranks split the field
line in `y` (`NYPE = np`). A correct component must give a decomposition-
independent result. We check two things:

1. the geometry output fields (`fieldline_geometry_*`) are identical across ranks;
2. the evolved solution (`Ne`, `Te`, …) matches the single-rank run to within the
   solver's own convergence floor (the 1D-threshold case sets that floor).

> **Prerequisites:** Docker (with `docker compose`), and a clone of hermes-3
> checked out on the `geometry_1d` branch. The published `:latest` image is built
> from `master`, so we build the branch code ourselves via the `work/hermes-3`
> override (steps 1–2).
>
> Multi-rank running (`run.sh hermes <case> <n>`) needs the parallel `run` mode
> that landed in #616, so make sure your image is current:
> `docker pull ghcr.io/boutproject/hermes-3:latest`. (If `./run.sh hermes
> work/case 4` logs `Processor number: 0 of 1`, your local image predates it.)
>
> All container commands go through the `docker/run.sh` helper, which wraps
> `docker compose`. It must be invoked from the `docker/` directory (it looks for
> `docker-compose.yaml` in the current directory), so every block below does
> `cd "$HERMES/docker"` first.

Pick your clone location once; every block below uses `$HERMES`:

```bash
export HERMES=$(git -C . rev-parse --show-toplevel)   # run from inside your clone
# ...or set it explicitly, e.g.  export HERMES=$HOME/hermes-3
cd "$HERMES" && git checkout geometry_1d
```

---

## 0. One-time setup

```bash
cd "$HERMES/docker"
sh setup.sh          # creates .env and the work/ mount (idempotent)
```

## 1. Stage the branch source into the build override

The image builds from `work/hermes-3` if present. Sync your checked-out branch
(and the googletest submodule, which the build needs for the unit tests) into it:

```bash
cd "$HERMES"
git submodule update --init external/googletest      # if not already populated
rsync -a --delete --exclude build/ --exclude .git/ ./ docker/work/hermes-3/
```

## 2. Build hermes-3 (with unit tests) in the container

Build **against the BOUT++ already compiled inside the image**
(`/hermes_project/build/boutpp-build`) — the branch only touches hermes-3
source, so there's no need to rebuild BOUT++. `HERMES_BUILD_BOUT=OFF` and
`HERMES_UPDATE_GIT_SUBMODULE=OFF` stop CMake from trying to re-fetch/rebuild the
bundled BOUT++ submodule (which fails, since we rsync'd without `.git/`):

```bash
cd "$HERMES/docker"
./run.sh shell bash -c '
  cmake -S /hermes_project/work/hermes-3 -B /hermes_project/work/hermes-3-build \
        -DCMAKE_PREFIX_PATH=/hermes_project/build/boutpp-build \
        -DHERMES_BUILD_BOUT=OFF -DHERMES_UPDATE_GIT_SUBMODULE=OFF \
        -DHERMES_TESTS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo &&
  cmake --build /hermes_project/work/hermes-3-build -j 4'
```

The executable lands at `/hermes_project/work/hermes-3-build/hermes-3`
(host: `docker/work/hermes-3-build/hermes-3`). The image's `run` mode picks up
this `work/hermes-3-build` override automatically.

## 3. Create the test cases (serial + parallel copies)

Start from the 1D-threshold case and add a **non-trivial** `[fieldline_geometry]`
section so the `{lpar}`-dependent expressions actually vary along the field line
(a constant geometry would pass trivially).

```bash
cd "$HERMES/docker/work"
for n in np1 np4; do
  rm -rf fieldline-par-$n
  cp -R "$HERMES/examples/tokamak-1D/1D-threshold" fieldline-par-$n
done
```

Two edits are needed in **both** `fieldline-par-np1/BOUT.inp` and
`fieldline-par-np4/BOUT.inp`.

**Edit 1 — enable the component.** In the `[hermes]` section, the 1D-threshold case has:

```ini
components = (d+, d, e,
              sheath_boundary_simple, braginskii_collisions, braginskii_friction,
              braginskii_heat_exchange, reactions, electron_force_balance,
              neutral_parallel_diffusion, braginskii_conduction, recycling)
```

Add `fieldline_geometry` as the **first** entry (component order matters — it must
set up the metric before the other components run):

```ini
components = (fieldline_geometry,
              d+, d, e,
              sheath_boundary_simple, braginskii_collisions, braginskii_friction,
              braginskii_heat_exchange, reactions, electron_force_balance,
              neutral_parallel_diffusion, braginskii_conduction, recycling)
```

**Edit 2 — configure it.** Append this section to the end of the file:

```ini
[fieldline_geometry]
# Non-trivial formulation to exercise the {lpar}-dependent expressions,
# which depend on the parallel-accumulated lpar being correct across ranks.
# mesh:length and mesh:length_xpt are already defined in the 1D-threshold case's
# [mesh] section, so no changes are needed there.
lpar_X1 = clamp( (mesh:length_xpt - {lpar}) / mesh:length_xpt, 0, 1)
lpar_X2 = clamp( ({lpar} - mesh:length_xpt) / (mesh:length - mesh:length_xpt), 0, 1)

lambda_q   = 1e-3
b          = 2.0
lambda_int = lambda_q * (1 + lpar_X2 * b)

poloidal_magnetic_field          = 1.0
upstream_toroidal_magnetic_field = 3.0
fieldline_radius = 0.5 + where({lpar} < mesh:length_xpt, 1 - cos(lpar_X1 * pi)/2, 1 - cos(lpar_X2 * pi)/2)

compute_Btor_from_R = true
diagnose            = true
```

That's all — the 1D-threshold case already sets `J = 1` in `[mesh]` (which
`fieldline_geometry` overrides) and defines `length`/`length_xpt`, so the
expressions above resolve without any further edits.

The stock example has `ny = 400`, which divides by 4, so np4 gives 100 interior
cells/rank. (If you change the rank count, use one that divides `ny`.)

## 4. Run at 1 and 4 ranks

`run.sh hermes <case> [n]` runs a case; the optional `n` is the MPI rank count
(default 1). It goes through the image's `run` mode, which launches `mpirun -np n`
(adding `--oversubscribe` automatically if `n` exceeds the cores Docker exposes):

```bash
cd "$HERMES/docker"
./run.sh hermes work/fieldline-par-np1 1
./run.sh hermes work/fieldline-par-np4 4
```

A correct 4-rank run logs `Running Hermes-3 in ... on 4 processes with mpirun`,
`Processor number: 0 of 4`, `NYPE=4`, and writes `BOUT.dmp.0.nc … BOUT.dmp.3.nc`
on the host under `docker/work/fieldline-par-np4/`.

> **Note — extra BOUT++ options.** The `run` mode only passes `-d <case>` to
> hermes-3; it does **not** forward trailing options such as `nout=10`. Put any
> such settings in the case's `BOUT.inp` instead (see the control in step 6).

## 5. Compare the runs

The image has `numpy` and `ncdump` but not xarray/netCDF4, so this small script
parses `ncdump` output. Save it as `docker/work/compare.py`:

```python
import subprocess, re, sys, glob, os
import numpy as np

def read_var(nc, var):
    out = subprocess.check_output(["ncdump", "-v", var, nc], text=True)
    m = re.search(re.escape(var) + r"\s*=\s*(.*?);", out, re.S)
    return np.array([float(x) for x in re.findall(r"[-+0-9.eE]+", m.group(1))])

def ydim(nc):
    out = subprocess.check_output(["ncdump", "-h", nc], text=True)
    return int(re.search(r"\by = (\d+)", out).group(1))

def field3d_vars(nc):
    # Match both time-evolving fields (t, x, y, z) and time-independent ones
    # (x, y, z) — the geometry outputs are constant in time and stored without
    # a time dimension.
    out = subprocess.check_output(["ncdump", "-h", nc], text=True)
    return re.findall(r"(?:double|float) (\w+)\((?:t, )?x, y, z\)", out)

serial_dir, par_dir = sys.argv[1], sys.argv[2]
f1 = os.path.join(serial_dir, "BOUT.dmp.0.nc")
fp = sorted(glob.glob(os.path.join(par_dir, "BOUT.dmp.*.nc")))
nf = len(fp)
assert nf > 1, "parallel dir must have >1 output file"

# Infer physical ny and guard width from the dumps (no hard-coded 400 / MYG=2):
#   y_full = NY + 2*MYG ; y_loc = NY/nf + 2*MYG
y_full, y_loc = ydim(f1), ydim(fp[0])
NY = round((y_full - y_loc) * nf / (nf - 1))
MYG = (y_full - NY) // 2
per = y_loc - 2 * MYG
print(f"serial y={y_full}  parallel y/file={y_loc}  nfiles={nf}  "
      f"=> NY={NY} MYG={MYG} interior/rank={per}")

allvars = field3d_vars(f1)
geom = [v for v in allvars if v.startswith("fieldline_geometry_")]
phys = [v for v in allvars if not v.startswith("fieldline_geometry_")]

def compare(var):
    r1 = read_var(f1, var); a1 = r1.reshape(r1.size // y_full, y_full)[-1]
    interior1 = a1[MYG:MYG + NY]
    parts = []
    for f in fp:
        r = read_var(f, var); a = r.reshape(r.size // y_loc, y_loc)[-1]
        parts.append(a[MYG:MYG + per])
    interior_p = np.concatenate(parts)
    diff = np.max(np.abs(interior1 - interior_p))
    rel = diff / (np.max(np.abs(interior1)) + 1e-300)
    return diff, rel

print("\n--- geometry outputs (must be machine-precision identical) ---")
gworst = 0.0
for v in geom:
    d, r = compare(v); gworst = max(gworst, r)
    print(f"  {v:38s} max|Δ|={d:.3e}  rel={r:.3e}")
print("  =>", "MATCH" if gworst < 1e-10 else f"MISMATCH (worst rel={gworst:.3e})")

print("\n--- evolved fields (compare against the 1D-threshold baseline) ---")
pworst, pv = 0.0, None
for v in phys:
    d, r = compare(v)
    if r > pworst: pworst, pv = r, v
    print(f"  {v:28s} max|Δ|={d:.3e}  rel={r:.3e}")
print(f"  => worst rel diff = {pworst:.3e} in {pv}")
```

Run it:

```bash
cd "$HERMES/docker"
./run.sh shell bash -c \
  "cd /hermes_project/work && python3 compare.py fieldline-par-np1 fieldline-par-np4"
```

The geometry outputs are the key correctness check — they come out
machine-precision identical across the 1- and 4-rank decompositions:

```
--- geometry outputs (must be machine-precision identical) ---
  fieldline_geometry_Bpxy                max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_Btxy                max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_Bxy                 max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_Rxy                 max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_cell_side_area      max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_cell_volume         max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_dlpol               max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_f_R                 max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_flux_tube_width     max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_lambda_int          max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_lpar                max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_magnetic_pitch      max|Δ|=0.000e+00  rel=0.000e+00
  fieldline_geometry_transport_broadening max|Δ|=0.000e+00  rel=0.000e+00
  => MATCH
```

## 6. (Recommended) Baseline control — is any residual difference geometry-specific?

The evolved-field comparison still shows an np1-vs-np4 difference (order ~1% and
above for some fields; worst rel `1.088e+00` in `SPe`). That is the `beuler`
steady-state solver's convergence floor, **not** a geometry bug. Prove it by
running the identical case with `fieldline_geometry` removed:

```bash
cd "$HERMES/docker/work"
for n in np1 np4; do
  rm -rf ctrl-$n && cp -R fieldline-par-$n ctrl-$n && rm -f ctrl-$n/BOUT.dmp.*.nc
  # then edit ctrl-$n/BOUT.inp: remove "fieldline_geometry, " from the components line
done
```

The `[fieldline_geometry]` section is now dead config. Since `run` mode doesn't
forward `input:error_on_unused_options=false` on the command line, disable the
unused-option check **inside** each `ctrl-*/BOUT.inp` instead — append:

```ini
[input]
error_on_unused_options = false
```

Then run and compare:

```bash
cd "$HERMES/docker"
./run.sh hermes work/ctrl-np1 1
./run.sh hermes work/ctrl-np4 4
./run.sh shell bash -c \
  "cd /hermes_project/work && python3 compare.py ctrl-np1 ctrl-np4"
```

The control shows an np-difference of the **same magnitude** on the evolved
fields — worst rel `6.507e-1` (in `snes_local_residual`), with the bulk fields
(`Ne`, `Nd`, `Te`, `Pd` …) all ~4–5%. That's comparable to the
`fieldline_geometry` case, confirming the residual is solver convergence, not
decomposition. (The large *relative* numbers are dominated by near-zero fields
where a tiny absolute difference blows up the ratio.)

## 7. Unit tests

```bash
cd "$HERMES/docker"
./run.sh shell bash -c \
  "/hermes_project/work/hermes-3-build/hermes_unit_tests --gtest_filter='*Fieldline*'"
```

Expected:

```
[==========] 11 tests from 1 test suite ran. (15 ms total)
[  PASSED  ] 11 tests.
```

---

## Gotchas

- **Run `./run.sh` from `docker/`, not `docker/work/`.** It (and the underlying
  `docker compose`) mounts `${PWD}/work`; if `PWD` is already `work/` you get
  `work/work` and files "disappear" inside the container.
- **`docker pull` first for multi-rank runs.** The parallel `run` mode arrived in
  #616; an older cached image runs `./run.sh hermes work/case 4` on a single
  process (`Processor number: 0 of 1`). `docker pull ghcr.io/boutproject/hermes-3:latest`.
- **`run` mode only passes `-d <case>` to hermes-3.** Trailing BOUT++ options
  (e.g. `nout=10`) are not forwarded — set them in `BOUT.inp`.
- **Don't leave a run in a foreground shell that you then kill.** `docker compose
  run` may leave the hermes container running detached; a second run writing the
  same `BOUT.dmp.*.nc` then fails with a NetCDF `Permission denied` (HDF5 file
  lock). `docker compose down --remove-orphans` clears it.

tbody-cfs:
@mikekryjak rebased onto latest main. Think this is good to go, and would be great to get it in while it's fresh

src/fieldline_geometry.cxx
mikekryjak: I talked to Lloyd and he will raise a separate PR to connect the reservoirs up.

mikekryjak (approved):
Good to go - thanks @tbody-cfs!

PR #235, opened 2024-07-22
Allow to compile custom BOUT++ directory

dschwoerer:
It can be convinient if you want your own version of BOUT++, but want to compile both BOUT++ and hermes for different architectures / os or with different flags.
This allows to have only one source for BOUT++ and on for hermes-3. You only need to invoke cmake and make/ninja just once per build and get always updates from BOUT++ and hermes-3 code changes.

An example cmake invocation can now look like this: 
```bash
cmake -S . -B build-opt -DHERMES_BUILD_BOUT=ON -DHERMES_UPDATE_GIT_SUBMODULE=OFF -DHERMES_TESTS=OFF -DHERMES_UNIT_TESTS=OFF -DHERMES_BOUT_SRC=$HOME/soft/BOUT-dev -DBOUT_ENABLE_METRIC_3D=ON -DBOUT_ENABLE_PYTHON=OFF -DBOUT_USE_FFTW=OFF -DBOUT_UPDATE_GIT_SUBMODULE=OFF -DBOUT_USE_PETSC=ON -DBOUT_USE_SUNDIALS=ON -DCHECK=0 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_FLAGS_RELWITHDEBINFO="-O3 -march=native -g -DNDEBUG"
```

or for a build with debugging:
```bash
cmake -S . -B build-debug  -DHERMES_BUILD_BOUT=ON -DHERMES_UPDATE_GIT_SUBMODULE=OFF  -DHERMES_TESTS=OFF -DHERMES_UNIT_TESTS=OFF -DHERMES_BOUT_SRC=$HOME/soft/BOUT-dev -DBOUT_ENABLE_METRIC_3D=ON -DBOUT_ENABLE_PYTHON=OFF -DBOUT_USE_FFTW=OFF -DBOUT_UPDATE_GIT_SUBMODULE=OFF -DBOUT_USE_PETSC=ON -DBOUT_USE_SUNDIALS=ON -DCHECK=4 -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_FLAGS_RELWITHDEBINFO="-Og -march=native -g -DNDEBUG"
```

This ensures that BOUT++ and hermes-3 also use the same flags for compilation, which makes it easier to check the impact of such changes.


bendudson:
Oddly the default option value is not being used, and instead the default default (?) 'OFF' is used
```
CMake Error at CMakeLists.txt:47 (add_subdirectory):
  add_subdirectory given source "OFF" which is not an existing directory.
```
Is it a particularly old version of cmake?

dschwoerer:
My bad, I should have tested more thoroughly.
`option` is ignored if the value is set, thus I have not noticed.
`option` is only for bool, thus the string is converted to a boolean, and apparently that folder is false.

bendudson (approved):

PR #238, opened 2024-08-02
Fixing kappa_par conversion

PoloidalLloyd:
Added Tnorm to kappa_par conversion in evolve_pressure (currently line https://github.com/bendudson/hermes-3/blob/8c450228e61d66200de225c168aa80fdfd91f7b2/src/evolve_pressure.cxx#L435)

The kappa_par conversion in evolve pressure was missing a factor of 1/Tnorm, causing the resulting conductivity to be too high. For the 1D recycling case Tnorm = 100 eV, which meant that calculated heat flux would be two orders of magnitude higher than supplied to the system via the pressure source. I've plotted this for the electron parallel conducitve heat flux (see below):

image

The sharp eyed may see that the electron heat flux is increasing between the x point and the target, above the supplied power, this is mirrored exactly by the ion heat flux, so energy is still conserved.

### DIFF
--- a/src/evolve_pressure.cxx
+++ b/src/evolve_pressure.cxx
@@ -432,7 +432,7 @@ void EvolvePressure::outputVars(Options& state) {
       set_with_attrs(state[std::string("kappa_par_") + name], kappa_par,
                      {{"time_dimension", "t"},
                       {"units", "W / m / eV"},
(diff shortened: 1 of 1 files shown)


bendudson (approved):
Thanks @PoloidalLloyd !

PR #242, opened 2024-08-07
collisions: Disable ion-neutral and electron-neutral by default

### LINKED ISSUE #241: Need to disable ion-neutral collisions by default
Ion-neutral collisions are currently enabled by default, but their effect is already included in the AMJUEL charge exchange rate we're using. This means enabling it will overpredict i-n exchange channels (not sure by how much).

I propose we simply make ion-neutral collisions disabled by default.

bendudson:
These may be double-counted if atomic reactions are included. Added a note to the manual.

ion-electron, electron-electron, ion-ion, and neutral-neutral collisions are still enabled by default.

Fixes #241.

### DIFF
--- a/docs/sphinx/components.rst
+++ b/docs/sphinx/components.rst
@@ -969,7 +969,8 @@ listed after all the species groups in the component list, so that all
 the species are present in the state.
 
 One of the most important is the `collisions`_ component. This sets collision
-times for all species, which are then used 
+times for all species, which are then used in other components to calculate
+quantities like heat diffusivities and viscosity closures.
 
 .. _sound_speed:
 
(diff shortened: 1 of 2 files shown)


mikekryjak:
How come you disabled `e-n` as well? Is that double counting something too?

bendudson:
I don't think e-n are double counted, but I thought this might be the least surprising set of defaults.
electron-neutral collisions aren't included in e.g Braginskii rates, so if someone is comparing to Braginskii then they would need to remember to disable e-n collisions. This means that all plasma-neutral interactions must be explicitly turned on, which might be a simple rule to remember.

The electron-driven reactions don't modify the electron collision frequency (https://github.com/bendudson/hermes-3/blob/master/include/amjuel_reaction.hxx#L68).

Happy to change this default, if something else would be easier for people to work with.

mikekryjak:
Hi @bendudson okay, I agree with that. There is a related question of what we do with the Braginskii collisionality implementation - whether we set Braginskii as default or not. I think that default should be Braginskii, and documentation and possibly a warning should warn the user if they are using multiple ions. I asked the question here: https://github.com/bendudson/hermes-3/pull/195

PR #248, opened 2024-08-16
WIP: Multiple improvements to sheath boundary and energy balance

bendudson:
Working on power balance diagnostics in 1D test case. 

MK edit:
This one turned into quite a journey into resolving an energy imbalance which went as high as 20% for an attached case. There are several fixes, as well as improvements that minimise zigzags at the boundary (a phenomenon that all codes struggle with if they don't have staggered grids) and new diagnostics. The diagnostics allowed us to reach a milestone - the first proper total energy flux balance.

Summary of changes:
- https://github.com/bendudson/hermes-3/pull/248/commits/e371fc7e6aac6d4d915c06a3f57897a627ba63f0: Viscous heating due to the diffusion of kinetic energy caused by the Lax flux viscosity
- https://github.com/bendudson/hermes-3/pull/248/commits/50bff348e455b607c2a1fd7df6cf60bdc84232b7 and https://github.com/bendudson/hermes-3/pull/248/commits/a9223e8abf089026f9e81699ebb53dcde4afd140: The ion bit in `sheath_boundary_simple` was set to `Vi>=Cs`, but it was `Vi=Cs` in the sheath current and electron loops. This would've caused issues in supersonic sheaths.
- https://github.com/bendudson/hermes-3/pull/248/commits/dde009a74eb10cb345341c523f11a84f49d9a2e8: New flag to keep sheath heat transfer but prevent particles leaving the model. This is super useful for debugging as it allows physical conduction-limited conditions with no neutrals.
- https://github.com/bendudson/hermes-3/pull/248/commits/97a21fdfaeb5a1672a57d5dbb4261145bb195566: This caused the majority of our energy imbalance. Pe had parallelSlices inside it due to a missing line. parallelSlices are 3D parallel interpolations that live on cell boundaries. This was being picked up instead of the sheath boundary extrapolation, which is wrong.
- https://github.com/bendudson/hermes-3/pull/248/commits/a002f1aaf3a85c13f2472c81ed2b4f52455076fd: Turns out when you add just a source of density and not pressure, this actually causes an energy sink: this is because when we add `N`, we have terms to dilute the total energy $\frac{5}{2}N_i T_i + \frac{1}{2} m_i N_i V_i^2$. This is done in the pressure equation by a factor proportional to `P=N*T` and in the momentum equation by a factor proportional to `NVi = N_i * V_i`. But because kinetic energy is $V^2$, it ends up reducing more than intended. There is now a new correction term for this called `numerical_viscous_heating` which you can turn on or off.
- https://github.com/bendudson/hermes-3/pull/248/commits/0009f3f9f7a720893894f4c576d90acbc2278b1e: We can either set the sheath boundary values directly or let the Lax flux drive to them. If we do the latter, previously it targetted a flux. It turns out if you make it target a velocity instead, it causes fewer zigzags near the boundary. This makes sense as it's more consistent.
- https://github.com/bendudson/hermes-3/pull/248/commits/ff0349b53a5b15c5284c868f327b808fa14f10d4: By default, our sheath extrapolation was set to Neumann for increasing quantities and exponential free for decreasing quantities. This caused an inconsistency if temperature was decreasing but N and P were increasing into the boundary, which can happen. This commit adds an option to force free extrapolation. We also found that linear extrapolation slightly reduces the boundary zigzag, presumably because it's more consistent with the equations (second order one sided differencing at the wall reduces to a linear extrapolation). Caution - this can make density negative upon extrapolation. An exponential extrapolation prevents this. **The default has been changed to exponential extrapolation**.

This work is part of my comparison to ReMKiT1D.  Here is a comparison showing that a free extrapolation has the best match to Remkit (note - cell centre plot only and **no neutrals**):
image

Testing on 1D-recycling with neutrals enabled, we found that all of the myriad fixes actually increased oscillations at the boundary unless we let the Lax flux set our Bohm condition (this is enabled by default, you can disable it with `fix_sheath_boundary_flux`):
image

That last result has the following flux balance:
image

A perfect balance would see the top series (Total Flow) stay constant. It doesn't quite manage that due to the "squiggles" in momentum at the boundary. The next stage is to explore a suggestion from @tbody-cfs which he implemented in GRILLIX (https://github.com/bendudson/hermes-3/pull/257). For now we will leave this as-is.

Remaining actions:

- [x] Rename flow diagnostics
- [x] Decide whether `numerical_viscous_flux` should be on or off by default, and whether `fix_boundary_flux` should be tied to it (see commit comment)
- [x] Add documentation on `sheath_boundary_simple` 
- [x]  Change the default extrapolation to exponential extrapolation (no Neumann when quantity is increasing)

### DIFF
--- a/src/evolve_pressure.cxx
+++ b/src/evolve_pressure.cxx
@@ -280,12 +280,16 @@ void EvolvePressure::finally(const Options& state) {
       ddt(P) += (2. / 3) * V * bracket(P, Apar_flutter, BRACKET_ARAKAWA);
     }
 
-    if (numerical_viscous_heating) {
+    if (numerical_viscous_heating || diagnose) {
       // Viscous heating coming from numerical viscosity
       Field3D Nlim = floor(N, density_floor);
       const BoutReal AA = get<BoutReal>(species["AA"]); // Atomic mass
-      Sp_nvh = (2. / 3) * AA * FV::Div_par_fvv_heating(Nlim, V, fastest_wave, fix_momentum_boundary_flux);
-      ddt(P) += Sp_nvh;
+      Sp_nvh = (2. / 3) * AA * FV::Div_par_fvv_heating(Nlim, V, fastest_wave, flow_ylow_kinetic, fix_momentum_boundary_flux);
+      flow_ylow_kinetic *= AA;
+      flow_ylow += flow_ylow_kinetic;
+      if (numerical_viscous_heating) {
+        ddt(P) += Sp_nvh;
+      }
     }
   }
 
@@ -539,7 +543,16 @@ void EvolvePressure::outputVars(Options& state) {
                     {"units", "W"},
                     {"conversion", rho_s0 * SQ(rho_s0) * Pnorm * Omega_ci},
                     {"standard_name", "power"},
-                    {"long_name", name + " power through Y cell face. Note: May be incomplete."},
+                    {"long_name", name + " conduction through Y cell face. Note: May be incomplete."},
+                    {"species", name},
+                    {"source", "evolve_pressure"}});
+
+      set_with_attrs(state[std::string("KineticFlow_") + name + std::string("_ylow")], flow_ylow_kinetic,
+                   {{"time_dimension", "t"},
+                    {"units", "W"},
+                    {"conversion", rho_s0 * SQ(rho_s0) * Pnorm * Omega_ci},
+                    {"standard_name", "power"},
+                    {"long_name", name + " kinetic energy flow through Y cell face. Note: May be incomplete."},
                     {"species", name},
                     {"source", "evolve_pressure"}});
     }
--- a/src/evolve_pressure.cxx
+++ b/src/evolve_pressure.cxx
@@ -247,7 +247,7 @@ void EvolvePressure::finally(const Options& state) {
 
     if (p_div_v) {
       // Use the P * Div(V) form
-      ddt(P) -= FV::Div_par_mod<hermes::Limiter>(P, V, fastest_wave);
+      ddt(P) -= FV::Div_par_mod<hermes::Limiter>(P, V, fastest_wave, flow_ylow);
 
       // Work done. This balances energetically a term in the momentum equation
       ddt(P) -= (2. / 3) * Pfloor * Div_par(V);
@@ -257,10 +257,11 @@ void EvolvePressure::finally(const Options& state) {
       // Note: A mixed form has been tried (on 1D neon example)
       //       -(4/3)*FV::Div_par(P,V) + (1/3)*(V * Grad_par(P) - P * Div_par(V))
       //       Caused heating of charged species near sheath like p_div_v
-      ddt(P) -= (5. / 3) * FV::Div_par_mod<hermes::Limiter>(P, V, fastest_wave);
+      ddt(P) -= (5. / 3) * FV::Div_par_mod<hermes::Limiter>(P, V, fastest_wave, flow_ylow);
 
       ddt(P) += (2. / 3) * V * Grad_par(P);
     }
+    flow_ylow *= 5. / 2; // Energy flow
 
     if (state.isSection("fields") and state["fields"].isSet("Apar_flutter")) {
       // Magnetic flutter term
@@ -345,7 +346,9 @@ void EvolvePressure::finally(const Options& state) {
 
     // Note: Flux through boundary turned off, because sheath heat flux
     // is calculated and removed separately
-    ddt(P) += (2. / 3) * FV::Div_par_K_Grad_par(kappa_par, T, false);
+    Field3D flow_ylow_conduction;
+    ddt(P) += (2. / 3) * Div_par_K_Grad_par_mod(kappa_par, T, flow_ylow_conduction, false);
+    flow_ylow += flow_ylow_conduction;
 
     if (state.isSection("fields") and state["fields"].isSet("Apar_flutter")) {
       // Magnetic flutter term. The operator splits into 4 pieces:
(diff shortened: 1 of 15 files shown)


bendudson:
With commit [0009f3f](https://github.com/bendudson/hermes-3/pull/248/commits/0009f3f9f7a720893894f4c576d90acbc2278b1e) the power balance in 1D without neutrals is now looking quite nice. 

The integrated input source (blue dashed) and total power flow (solid pink) are on top of each other:
image
[Note the huge power: It's 300MW because the cross-section area of the flux tube is 2pi square meters]

```
Total input power:     313.2444925223502 MW
  |- Ion heating:      156.6222462611751 MW
  |- Electron heating: 156.6222462611751 MW

Total power loss: 313.13056169921424 MW
  |- Ions:              178.27349896833005 MW
      |- Convection          127.33821354880719 MW
      |- Kinetic energy      54.98127756518318 MW
      |- Conduction          -4.0459921456603105 MW
  |- Electrons:         134.85706273088422 MW
      |- Convection          74.92065741132512 MW
      |- Kinetic energy      0.01497311480533311 MW
      |- Conduction          59.92143220475376 MW
```

mikekryjak:
That looks great, in particular the change from damping flux to velocity was inspired!
One thing jumps out at me in that plot is that ion KE is negative towards the target. Is this 1D-recycling with no neutrals and no sheath flow? 
If it's steady state, then there shouldn't be a particle source upstream, and therefore there shouldn't be any flows anywhere... I think. Maybe it's not perfectly steady state? It takes a while to equilibrate.

bendudson:
I think the ion heat conduction (green) is slightly negative because the sheath gamma_i is lower than the sum of KE + internal energy advection into the sheath. There's therefore a source of heat at the sheath (4 MW) that is conducted into the plasma.

src/evolve_pressure.cxx
```diff
@@ -512,6 +538,35 @@ void EvolvePressure::outputVars(Options& state) {
...
+                    {"units", "W"},
+                    {"conversion", rho_s0 * SQ(rho_s0) * Pnorm * Omega_ci},
+                    {"standard_name", "power"},
+                    {"long_name", name + " kinetic energy flow through Y cell face. Note: May be incomplete."},
+                    {"species", name},
+                    {"source", "evolve_pressure"}});
+    }
+
+    if (numerical_viscous_heating) {
+      set_with_attrs(state[std::string("SP") + name + std::string("_nvh")], Sp_nvh,
```
mikekryjak, line 562: To keep with the most common naming convention, this should be converted to energy and named E<species>_nvh.

mikekryjak: Unless we make the naming convention so that e.g. Ed+_nvh would be for terms relevant to balance or physics post-processing, where most variables are already in the form of energy, and e.g. `SPd+_nvh` would specifically refer to equation terms. The problem is that `S` already means density source.

dschwoerer: I just tested: `fmt::format("SP{}_nvh", name)` is around 40% faster then `std::string("SP") + name + std::string("_nvh")` :-)
<details>

```
[dave@raven04: ~]👍 cat /viper/u/dave/soft/hermes-2/relax/test.cc
#include <string>
#include <fmt/format.h>

int main() {
  const std::string name{"h+"};
  for (int i=0 ; i< 10000000; ++i) {
#ifdef FMT
    const auto fu = fmt::format("SP{}_nvh", name);
#else
    const auto fu = std::string("SP") + name + std::string("_nvh");
#endif
    asm("");
  }
}
[dave@raven04: ~]👍 g++ /viper/u/dave/soft/hermes-2/relax/test.cc -lfmt && time ./a.out

real    0m1.011s
user    0m1.010s
sys     0m0.000s
[dave@raven04: ~]👍 g++ -DFMT /viper/u/dave/soft/hermes-2/relax/test.cc -lfmt && time ./a.out

real    0m0.669s
user    0m0.665s
sys     0m0.004s
```
</details>

mikekryjak:
@bendudson I propose we rename the flux diagnostics to be more consistent with the rest of the naming scheme and a bit clearer. I was thinking of:

Ion particle flux: pfd+_ylow
Ion conduction flux: efd+_cond_ylow
Ion energy advection flux: efd+_adv_ylow
Ion total energy flux: efd+_tot_ylow
Ion momentum advection flux: mfd+_adv_ylow
Ion viscosity flux: mfd+_visc_ylow

Thoughts?

src/evolve_pressure.cxx
```diff
@@ -148,6 +148,16 @@ EvolvePressure::EvolvePressure(std::string name, Options& alloptions, Solver* so
...
     .withDefault<bool>(false);
+
+  numerical_viscous_heating = options["numerical_viscous_heating"]
+    .doc("Include heating due to numerical viscosity?")
+    .withDefault<bool>(false);
+
+  if (numerical_viscous_heating) {
+    fix_momentum_boundary_flux = options["fix_momentum_boundary_flux"]
+      .doc("Fix Y boundary momentum flux to boundary midpoint value?")
+      .withDefault<bool>(false);
```
mikekryjak, line 159: @bendudson how come `fix_momentum_boundary_flux` is only available with `numerical_viscous_heating`? I only just noticed this. 

Second question - are there any downsides to having numerical viscous heating enabled? At the moment it defaults to disabled, which would lead to a small energy imbalance.

mikekryjak: It's there because in `evolve_pressure`, `fix_momentum_boundary_flux` is only relevant if `numerical_viscous_heating` is enabled, because you need the operator to calculate it.

mikekryjak:
@bendudson good to review/merge

PR #259, opened 2024-09-12
Update 1D-recycling example

mikekryjak:
- CVODE version now consistent with beuler
- Reduced CVODE `nout` to 100 as while 400 may be necessary to get rid of all the tiny oscillations but may confuse new users
- Reduced `d` density floor to prevent unphysical temperatures
- Enabled `diagnose=true` in all components
- Changed sheath BC to `sheath_boundary_simple` as it's the most thoroughly tested one

To do:

- [ ] Change gamma setting to something more expected (4.5 for electrons, 2.5 for ions?)
- [ ] Test again
- [ ] Change collisions to Braginskii
- [ ] Improve comments


bendudson (approved):
Thanks @mikekryjak !

mikekryjak:
Interestingly, `sound_speed` seems to slow down CVODE by about 10% - 20%. Probably because of the electrons being less damped now.

mikekryjak:
Scratch that, adding sound_speed doubled the simulation time. @bendudson what do you think we should do with this...? On one hand, adding the component increases accuracy, on the other hand it will slow people's runs down and may lead to crashes during transients. In the end we are way less dissipative than SOLPS and other codes anyway...


Also, I found that the odd neutral temperature profiles were due to the neutrals hitting the floor. Lowering the floor to 1e-7 resolves the issue. Although you get the funny bump in Tn which Stefan spent a good amount of time tracking down - it's due to neutral friction and only appears when you correctly include the 5/2 factor in the neutral diffusion advection, which I did for 1D some time ago.

image

mikekryjak:
Done this from scratch in https://github.com/boutproject/hermes-3/pull/314 with no PR dependencies.

PR #268, opened 2024-10-18
Coulomb log fix

PoloidalLloyd:
S. Mijin Identified a discontinuity in the NRL formulary fits for the electron ion coulomb log, this can be rectified by altering the change over point to exp(2.0) see attached. 

- [ ] Run tests on 1D recycling. 
image


bendudson (approved):
Thanks @PoloidalLloyd for dealing with this clog! And to Stefan for spotting the error in NRL.

PR #283, opened 2025-01-13
Reduce main.cxx for MMS test to the minimal code for a single operator

johnomotani:
Strips down the main.cxx file to a minimal amount of code. You probably want to restore David's original (except maybe the parallel boundary conditions, which aren't needed for non-FCI) eventually, but this should be more readable to start with.


mrhardman:
Superseded by #285.

PR #285, opened 2025-01-24
Nonorthogonal with minimal differential operators MMS test.

mrhardman:
Fixes to the `Div_a_Grad_perp_nonorthog(a, f)` operator and new automatic tests showing a correct implementation `hermes-3/tests/mms/orthogonal_test.py` and `hermes-3/tests/mms/nonorthogonal_test.py`. These tests can be readily extended to test new nonorthogonal operators and the other existing operators in the Hermes-3 code base.

See `/hermes-3/tests/mmsREADME.md` for some documentation. Comments and requests appreciated.

Thanks to @dschwoerer for providing an initial MMS test to develop.

Below is the plot output from the tests, showing the error (blue), the expected (orange), and the fit of the error (green). Increase the range of $`\Delta`$ by increasing `ntest` in the `test_input` dictionary in `hermes-3/tests/mms/orthogonal_test.py` and `hermes-3/tests/mms/nonorthogonal_test.py`. Note that since the number of grid points $`N_x = N_y = N_z`$, the cost of the test rapidly increases with `ntest`.

`FV::Div_a_Grad_perp(a, f)` with orthogonal metric
fig_0
`Div_a_Grad_perp_nonorthog(a, f)` with orthogonal metric
fig_1
`Div_a_Grad_perp_nonorthog(a, f)` with nonorthogonal metric
fig_0_non


dschwoerer:
Can you either change this PR to go into master or open a new one that goes into master:
https://github.com/bendudson/hermes-3/compare/bendudson:master...bendudson:nonorthogonal-divops-mms-test-minimal?expand=1

Like this it is hard to see the changes you introduced.

Unfortunately I cannot do that, as I am not a collaborator :-(

.pip_install_for_tests.sh
```diff
@@ -18,3 +18,16 @@ do
...
+git clone https://github.com/boutproject/xBOUT.git
+cd xBOUT
+pip3 install --user .
+cd ..
+
+# Install Hermes-3
+git clone https://github.com/boutproject/xhermes.git
+cd xhermes
+pip3 install --user .
+cd ..
```
dschwoerer, line 32: If you want to do this, I would suggest:
```suggestion
pip install  --user git+https://github.com/boutproject/xBOUT.git
pip install  --user git+https://github.com/boutproject/xhermes.git
```

But I think, if there are changes that are wanted / needed, there should be a release. Otherwise everybody will always start installing from git, which is imho not as nice, as using the released and version package.

Releases are cheap enough, so lets use them!

mrhardman: At the moment I need the latest master of XBOUT so that the grid loading could function properly in the test. This was my workaround, since I didn't know how to do that through pip. Happy to refactor as suggested if @bendudson @mikekryjak agree on the style change -- I didn't invent the xhermes line.

mrhardman:
Changed base to master in response to @dschwoerer's review.

bendudson (approved):
Amazing work. Thanks @mrhardman !

src/div_ops.cxx
```diff
@@ -951,6 +929,353 @@ const Field3D Div_a_Grad_perp_upwind(const Field3D& a, const Field3D& f) {
...
+  if (f.hasParallelSlices() && a.hasParallelSlices()) {
+    // Both inputs have yup and ydown
+
+    fup = f.yup();
+    fdown = f.ydown();
+
+    aup = a.yup();
+    adown = a.ydown();
+
+    mesh->communicate(fddx_xhigh);
```
dschwoerer, line 1000: Note that this does not work for fci.
for fci the right thing would be to do the above calculation also for the parallel slices.
```suggestion
    if (f.isFci()) {
        throw BoutException("This is not implemented for FCI.\n"
        "The above calculation needs to be done for the parallel slices as well");
    }
    mesh->communicate(fddx_xhigh);
```

src/div_ops.cxx
```diff
@@ -951,6 +929,353 @@ const Field3D Div_a_Grad_perp_upwind(const Field3D& a, const Field3D& f) {
...
+  }
+
+  if (bout::build::use_metric_3d) {
+    // 3D Metric, need yup/ydown fields.
+    // Requires previous communication of metrics
+    // -- should insert communication here?
+    if (!coord->g23.hasParallelSlices() || !coord->g_23.hasParallelSlices()
+        || !coord->dy.hasParallelSlices() || !coord->dz.hasParallelSlices()
+        || !coord->Bxy.hasParallelSlices() || !coord->J.hasParallelSlices()) {
+      throw BoutException("metrics have no yup/down: Maybe communicate in init?");
```
dschwoerer, line 1022: Metrics must not be communicated:
```suggestion
      throw BoutException("metrics have no yup/down");
```

src/div_ops.cxx
```diff
@@ -951,6 +929,353 @@ const Field3D Div_a_Grad_perp_upwind(const Field3D& a, const Field3D& f) {
...
+
+  // Flux in X
+
+  for (int i = mesh->xstart - 1; i <= mesh->xend; i++) {
+    for (int j = mesh->ystart; j <= mesh->yend; j++) {
+      for (int k = 0; k < mesh->LocalNz; k++) {
+        // Calculate flux from i to i+1
+
+        const int kp = (k + 1) % mesh->LocalNz;
+        const int km = (k - 1 + mesh->LocalNz) % mesh->LocalNz;
```
dschwoerer, line 1233: same as above

CMakeLists.txt
```diff
@@ -245,6 +245,29 @@ if(HERMES_TESTS)
...
+  
+# Method of manufactured solutions (MMS) tests
+  option(HERMES_DIVOPS_MMS_TESTS "Build the differential operator MMS tests" ON)
+  if(HERMES_DIVOPS_MMS_TESTS)
+    if(NOT PACKAGE_TESTS)
+      # Not running BOUT++ tests
+      # Since BOUT++ includes googletest, adding it again causes errors
+      set(CMAKE_CXX_STANDARD 14) # Needed to compile GoogleTest on Darwin
+      # add_subdirectory(external/googletestmms)
+    endif()
```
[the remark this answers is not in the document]
mrhardman: Agreed, will remove in a commit.

src/div_ops.cxx
```diff
@@ -70,8 +69,8 @@ const Field3D Div_par_diffusion_index(const Field3D &f, bool bndry_flux) {
 ////////////////////////////////////////////////////////////////////////////////////////////////
 // XPPM methods
 
-BoutReal BOUTMIN(const BoutReal &a, const BoutReal &b, const BoutReal &c,
-                 const BoutReal &d) {
+BoutReal BOUTMIN(const BoutReal& a, const BoutReal& b, const BoutReal& c,
+                 const BoutReal& d) {
```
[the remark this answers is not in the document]
mrhardman: This seems orthogonal to this PR -- can you please explain what the suggested change does here?

src/div_ops.cxx
```diff
@@ -70,8 +69,8 @@ const Field3D Div_par_diffusion_index(const Field3D &f, bool bndry_flux) {
...
   BoutReal r1 = (a < b) ? a : b;
   BoutReal r2 = (c < d) ? c : d;
   return (r1 < r2) ? r1 : r2;
```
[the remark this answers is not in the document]
mrhardman: This seems orthogonal to this PR -- can you please explain what the suggested change does here? I don't know for sure whether or not this line is covered by a test.

mrhardman:
@dschwoerer Could you please make a PR with your suggestions into this PR? It looks like some of the functions that you expect to be available are not.

.pip_install_for_tests.sh
mikekryjak: The style change seems equivalent - if that is the case then sure, why not.
@dschwoerer I agree with you with regards to releases. I've had practically zero time to contribute to xHermes for a while. We may be getting some RSE support that could help with this in the next two months.

mrhardman:
With commit https://github.com/boutproject/hermes-3/pull/285/commits/9ecddebf7a3e57536989df88ced3804a3c0323e9 we can perform a diffusion test solving an equation (up to normalisation) of 

```math
 \frac{\partial n}{\partial t} = \nabla \cdot \left( D_n \nabla_\perp n \right)
```
and we choose `D_n` to be constant. The input file used for this test is 

```
nout = 50
timestep = 50000.0

MZ = 1

[mesh]

file = "bout.grd.nc"

extrapolate_y = false

[mesh:paralleltransform]
type = shifted

[solver]
type = cvode
use_precon = true
mxstep = 1e5

mxorder = 3
atol = 1e-12
rtol = 1e-5

[hermes]
components = (d, classical_diffusion)

Nnorm = 1e19  # Reference density [m^-3]
Bnorm = 1   # Reference magnetic field [T]
Tnorm = 75   # Reference temperature [eV]
loadmetric = false

[classical_diffusion]
custom_D = 1.0
diagnose = true

##################################################################
# Neutrals

[d]
type = evolve_density
charge = 0
AA = 2

diagnose = true
[Nd]

function = 1 - 0.5*x^2 + x

bndry_core = dirichlet(1.0)  # Core boundary high density
bndry_all = neumann   # All other boundaries low density


```

The result at the last timestep is plotted here, showing that the solution is very close to 1.0e19, as required at long times by the boundary conditions.
Last_timestep_Nd
The grid can be generated by modifying the examples from the Hypnotoad integrated tests.

mrhardman:
> @mrhardman can you move the test to `tests/integrated/neutral_mixed_mms`? Thanks!

These tests in this PR are not of neutral mixed but of the differential operators. The tests of the neutral mixed module are in #327. See issue #378. At the moment both the differential operator tests and the neutral mixed mms tests in #327 must live in the same folder. Can we use a name like `tests/integrated/differential_operator_mms`?

mikekryjak (changes requested):
Looks good to me apart from two minor points:
- Let's rename the `mms` folder to `mms-operators` since there are other mms tests in other folders.
- `rnn_override` could use a better name.

PR #288, opened 2025-02-15
sheath_boundary: Add heat flux diagnostics

bendudson:
Saves total energy sink and gamma heat transmission factors.

### DIFF
--- a/include/sheath_boundary.hxx
+++ b/include/sheath_boundary.hxx
@@ -75,6 +75,9 @@ struct SheathBoundary : public Component {
   ///
   ///
   void transform(Options &state) override;
+
+  /// Save diagnostics
+  void outputVars(Options &state) override;
 private:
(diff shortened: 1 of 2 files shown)


mikekryjak:
@bendudson since `sheath-boundary` currently has several issues and is probably not conserving energy as it's missing the fixes we made for `sheath-boundary-simple` in https://github.com/bendudson/hermes-3/pull/248, I think before we modify `sheath-boundary`, we should merge the two components together. 

I think the only difference between `sheath-boundary` and `sheath-boundary-simple` is the Tskhakaya sheath gamma model, so we should implement that feature into `sheath-boundary-simple`, rename it to `sheath-boundary` and delete the old `sheath-boundary` component.

This is also important for @dschwoerer who is working on the sheath BC to add FCI compatibility.

PR #289, opened 2025-02-16
fixed-fraction-radiation: Disable core radiation by default

mikekryjak:
In fixed fraction impurity radiation models, it's assumed that the impurity fraction and cooling curve remain constant throughout the domain. In reality, the impurities in the core have a much longer residence time than ones in the edge, which reduces their radiation power to coronal levels. 

Based on conversations with David Moulton, this can be emulated by simply disabling core radiation. This can prevent runaway detachment fronts.

This PR includes the option `no_core_radiation` and sets it to **True** by default.

- [x] Implement and simple test
- [x] Test in production MAST-U case
- [x] Documentation

image
image


mikekryjak:
@bendudson ready to go

bendudson (approved):

PR #295, opened 2025-02-20
Unify docs theme with BOUT++

mikekryjak:
This PR changes the docs theme to be like BOUT++, see screenshot below. This theme is `sphinx_book_theme` and you must pip install it before building the docs locally. The theme is the same on Readthedocs as locally which helps to make the docs writing experience more predictable - we previously had a different theme locally and it made things look very different once they were up.

One thing I would appreciate feedback on is how to add `sphinx_book_theme` to some sort of python environment requirements so that users don't get confused. I don't think we give any instructions right now on how to build the documentation, so maybe we could add a documentation page on that and instruct the user to install the right theme there? Happy to do that if you think that's the right way forward.

Would also appreciate any tips on Readthedocs gotchas. I've never looked into it and I'm not sure if I'll be breaking anything here.

image

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 breathe
 recommonmark
-sphinx-rtd-theme
+sphinx-book-theme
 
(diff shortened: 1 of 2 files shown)


mikekryjak:
@bendudson ready to go. I just had to add the theme to readthedocs requirements.

mikekryjak:
Solves https://github.com/bendudson/hermes-3/issues/263

bendudson (approved):

PR #297, opened 2025-02-21
Neutral toroidal slip for 1D simulations

bendudson:
Allow parallel flow into the target, by setting the component of the velocity into the target to zero. This is a combination of the parallel flow and perpendicular flow.

Sets a Neumann boundary condition on the parallel velocity, and calculates the perpendicular velocity (i.e. neutral pressure gradient) at the target such that the net flow into the target is zero.

**MK comments**
Here is my understanding of how this boundary condition works:
- Usually, neutral velocity **must** be zero at the boundary in a fluid model if no neutrals are leaving the domain. There is no other way. It was tempting to assume that we could have a momentum source there to represent FC velocity towards the domain, but then no particles would ever get to the target in the first place to become reflected.
- The only way to relax this requirement is to allow some velocity "out of the page". Physically, picture a neutral bouncing off the target in a direction perpendicular to the field line and flying off into other field lines. It is effectively cross-field transport which enhances parallel diffusion through its poloidal component. 
- In 1D, this translates as parallel particle transport in the upstream direction. The way to get this to work in our model is to make this transport diffusive. This way you could have the momentum equation transport neutrals to the target, but then use the diffusion equation to transport them in the opposite direction after reflection - all simultaneously!
- This is achieved by the pressure diffusion model, like in 2D. To begin with, we can assume that all particles undergo fast reflection. This is equivalent to a "slip" condition with no losses at the target. We then calculate a neutral pressure at the target such that pressure diffusion occurs with a velocity equal and opposite to that solved by the neutral momentum at the target. 
- In practice, this is like a more physically meaningful neutral redistribution method than that in SD1D and SOL-KiT. It should lead to high flows in the neutral cloud and into the target, and transport the particles from near the target to farther upstream.


include/amjuel_reaction.hxx
mikekryjak, line 1: I'd like to make sure I fully understand this. How come using get<Field3D> prevents setting a velocity BC after the reactions? And what is the difference between get<Field3D> and GET_VALUE in general?

bendudson: Hey @mikekryjak ! The idea is to prevent quantities from being modified after they are used. When `get` is called it sets two attributes: "final" and "final-domain" 
https://github.com/bendudson/hermes-3/blob/master/include/component.hxx#L117
When `set` is called it checks if these attributes have been set:
https://github.com/bendudson/hermes-3/blob/master/include/component.hxx#L231

To allow quantities to be used in the domain before boundary conditions have been applied, there is `getNoBoundary` that only sets the "final-domain" attribute:
https://github.com/bendudson/hermes-3/blob/master/include/component.hxx#L183
This allows `setBoundary` to set the boundary because it only checks the "final" attribute:
https://github.com/bendudson/hermes-3/blob/master/include/component.hxx#L261
In this case the boundary condition on pressure depended on the collision frequency, and the collision frequency depended on the pressure (temperature) but only in the domain. When collisions are calculated it therefore should use `getNoBoundary`, which will allow a boundary to be applied later but prevent the values in the domain from being modified.

Both `get` and `getBoundary` have an optional second argument `location`. This is a string that is included in the error messages from `set` and `setBoundary`. These tell the user where the quantity was used; otherwise the user just has to guess which components are in the wrong order. The easiest way to set that string is to use `GET_VALUE` and `GET_NOBOUNDARY`. These combine the name of the file and the line number where they are run, passing this as the location to `get`:
https://github.com/bendudson/hermes-3/blob/master/include/component.hxx#L163

mikekryjak: Wow, I'm glad I'm learning about this! I think we need to put this in the docs under a new section called "Developer manual" or similar. I raised an issue: https://github.com/bendudson/hermes-3/issues/299

mikekryjak:
Bumping this as it's relevant to the ongoing work on our neutral model, especially the parallel momentum boundary conditions.

Issue #299, opened 2025-02-23
Need developer docs


mikekryjak:
We need to write some docs for developers. While some things are covered by the BOUT++ documentation, there is a lot of stuff in there which is not strictly relevant to Hermes-3, and it's not easy to navigate.

I think we need sections for some key concepts, like:

- [ ] GET_VALUE and SET_VALUE (https://github.com/bendudson/hermes-3/pull/297#discussion_r1966588803)
- [ ] BOUT_FOR loops and similar (how to loop over boundary, how to tell if you're in the core)
- [ ] How boundary conditions work (e.g. we set fundamental boundary conditions and then achieve what we want with sources)
- [ ] How `collisions` works
- [ ] What should be in the header file, what should be in the source file
- [ ] Adding new options
- [ ] Adding new post-processing variables
- [ ] Some minimal content of C++ tips, e.g. what's a pointer, what's a reference, classes, templates

mikekryjak:
This is basically all done now.

PR #301, opened 2025-03-05
Misc. improvements to output diagnostics

bendudson:


mikekryjak:
This is great, I forgot that post-processing variables will be saved just once if you don't include the time dimension. This should go into https://github.com/boutproject/hermes-3/issues/299.

mikekryjak:
@bendudson This one seems useful, was there a specific reason why it's unmerged?

PR #308, opened 2025-03-14
3D FCI support

dschwoerer:
The `sheath_boundary_simple.cxx` has not been updated for a bit, but this shows how the boundary code can be simplified.


include/sheath_boundary_parallel.hxx
```diff
@@ -0,0 +1,115 @@
...
+
+  bool always_set_phi; ///< Set phi field?
+
+  Field3D wall_potential; ///< Voltage at the wall. Normalised units.
+
+  bool floor_potential; ///< Apply floor to sheath potential?
+
+  bool legacy_match{true};
+
+  Field3D fromFieldAligned(const Field3D& f) {
```
bendudson, line 92: I think these to/fromFieldAligned functions should be renamed: The caller may expect that the result is transformed, but if using FCI then it is not. That affects whether the caller can treat `f(x, y+1, z)` as being along a field line from `f(x, y, z)`.

I'm coming to the opinion that trying to handle both yup/down and field-aligned cases with the same code is going to be fragile and easy to introduce errors. I think I would prefer a Mesh- or ParallelTransform-level boolean: Are we using yup/down fields for differencing or not?

Issue #303, opened 2025-03-07
Neutral density flooring causing large slow down


PoloidalLloyd:
In low density systems the neutrals are hitting the density floor, which leads to unphysical upstream neutral temperatures, and oscillations in temperature and density that are causing large slow downs (approximately 2 orders of magnitude) in 1D. This is shown in the gif below. Despite this bug being in 1D, this could potentially cause slow downs in 2D. 

![Image](https://github.com/user-attachments/assets/6f1d9e97-2c35-4d57-9efc-2eb69614e56c)

See attached the specific case below
[corrected_fcx_on_high_density_4.8e18_start_density_floor_issues!.zip](https://github.com/user-attachments/files/19130600/corrected_fcx_on_high_density_4.8e18_start_density_floor_issues.zip)

PR #314, opened 2025-03-30
Reduce default density floor, improve 1D examples and docs

mikekryjak:
Adds new 1D example called `1D-threshold` based on `1D-recycling`. 
Also reduces the default density floor from 1e-5 to 1e-7 for `evolve_density`, `evolve_energy`, `evolve_pressure` and `evolve_momentum`. This is because the default 1e-5 was hit in `1D-recycling` and can be easily hit in 2D simulations too, in particular for neutral density in the core region. 1e-7 corresponds to the default value of `atol` so it should be fine.

Changes:

- [x] Reduce density floors to prevent unphysically low neutral temperatures (due to this bug: https://github.com/boutproject/hermes-3/issues/303)
- [x] Change sheath BC to sheath_boundary_simple and expose gamma settings
- [x] Increase upstream density to take the conditions to detachment threshold where performance is much faster due to lower target temperatures
- [x] Improve comments 
- [x] Remove contents of `README.md` and replace with reference to documentation
- [x] Move all other 1D examples into a directory called `extra`. This is to indicate that they may be out of date or not work as expected.
- [x] Redo gif animation
- [x] Update documentation
- [x] Fix doc figure
- [x] Fix floor test fail


mikekryjak:
@bendudson ready for review

bendudson (approved):
Thanks @mikekryjak !

Issue #305, opened 2025-03-12
Crash if SOL/PFR recycling is disabled but pump enabled


mikekryjak:
This recent change made sure that recycling diagnostics are done per channel:
https://github.com/boutproject/hermes-3/commit/89adbfd4a14fa2095d08e4234b6b1aee6c010980

This moves the initialisation and allocation of the diagnostics into an if statement which is true if SOL/PFR recycling is enabled:
https://github.com/boutproject/hermes-3/blob/1e067b8690608eaf9ac845f1205d0e1bc24cbbef/src/recycling.cxx#L286-L289

Which causes a crash when SOL/PFR recycling is disabled but the pump is enabled:

> Monitor signalled to quit (exception set_with_attrs: Field3D assigned to Sd_pump is not allocated)
Error encountered: set_with_attrs: Field3D assigned to Sd_pump is not allocated

There needs to be an exception if this combination of options is chosen, because the pump uses SOL/PFR recycling machinery and so requires it to be enabled.

bendudson:
Fixed by #324

Issue #250, opened 2024-08-28
Radial recycling causes crash on cold start in 2D with core sources


mikekryjak:
I have just started a simulation on a new grid and wanted to do it from scratch. I was only able to start the simulation without core sources and with everything turned down/off, as per the last time I did this with the MAST-U cases.

I just did some more testing and found that the case can be started from cold with all the bells and whistles (core sources present, final density and pressure settings, etc) if radial recycling is disabled. 

This suggests there is likely a bug in `recycling.cxx` which causes an unphysical situation only at the very start of the solution.

mikekryjak:
The error message is as below and can be worked around by disabling SOL and PFR recycling:

> ====== Back trace ======
>  -> virtual void Recycling::transform(Options&) on line 148 of '/users/mjk557/scratch/hermes-3/src/recycling.cxx'
> 
> ====== Exception thrown ======
> Setting invalid value for 'species:d:density_source'

bendudson:
Closed by #324

PR #324, opened 2025-04-04
Add flooring to neutral pump 

mikekryjak:
Solves https://github.com/boutproject/hermes-3/issues/250
Solves https://github.com/boutproject/hermes-3/issues/305

There has been a bug for ages which causes a crash when starting a case with radial recycling from scratch. Turns out it's actually a problem with the pump. I put a free BC in there:

https://github.com/boutproject/hermes-3/blob/a33e1ba046a4090ac588dc53ab1ff0017515c0ff/src/recycling.cxx#L352-L358

This is a problem as neutrals are often initialised with zero density, causing a divide by zero and the error message:

> ====== Exception thrown ======
Setting invalid value for 'species:d:density_source'

I have now floored density, pressure and temperature:

https://github.com/boutproject/hermes-3/blob/d8232ba74430d92cb21fb462d90e300482c1161c/src/recycling.cxx#L175-L177

The floored values are only used in the calculations of sheath neutral quantities, nowhere else as I can't spot any other bits that would require it:

https://github.com/boutproject/hermes-3/blob/d8232ba74430d92cb21fb462d90e300482c1161c/src/recycling.cxx#L359-L370

### DIFF
--- a/src/recycling.cxx
+++ b/src/recycling.cxx
@@ -42,6 +42,9 @@ Recycling::Recycling(std::string name, Options& alloptions, Solver*) {
                          .doc("Name of the species to recycle into")
                          .as<std::string>();
 
+    density_floor = options["density_floor"].doc("Minimum density floor").withDefault(1e-7);
+    pressure_floor = density_floor * (1./get<BoutReal>(alloptions["units"]["eV"]));
+
     diagnose =
       from_options["diagnose"].doc("Save additional diagnostics?").withDefault<bool>(false);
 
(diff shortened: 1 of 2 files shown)


mikekryjak:
@bendudson ready for review. Not sure why the docs are failing to build, seems spurious.

src/recycling.cxx
```diff
@@ -353,13 +360,13 @@ void Recycling::transform(Options& state) {
               // These are NOT communicated back into state and will exist only in this component
               // This will prevent neutrals leaking through cross-field transport from neutral_mixed or other components
               // While enabling us to still calculate radial wall fluxes separately here
-              BoutReal nnguard = SQ(Nn[i]) / Nn[is];
-              BoutReal pnguard = SQ(Pn[i]) / Pn[is];
-              BoutReal tnguard = SQ(Tn[i]) / Tn[is];
+              BoutReal nnguard = SQ(Nnlim[i]) / Nnlim[is];
```
bendudson, line 363: Question here: If using `SQ(Nnlim)` / Nnlim then the guard value will never fall below the floor. If `Nn` is lower than `Nnlim` then nnguard will be greater than Nn. A possibility is to use `SQ(Nn[i]) / Nnlim[is]` so that the guard cell value goes to zero as `Nn` goes to zero.

bendudson (approved):

PR #330, opened 2025-04-18
Energy transfer diagnostics

bendudson:
Many optional diagnostic outputs, saving energy transfer channels for post-processing.

Not ready for merging: Need to consolidate sheath boundary diagnostics.


mikekryjak:
This looks great! One thing I noticed is that the diagnostic nomenclature is different than existing, e.g. `E_Div_Vpar_KE_d+` instead of `Ed+_Div_Vpar_KE`.

mikekryjak:
@bendudson are these the diagnostics you've used for the energy balance work? It would be good to get these merged, but I would prefer if we edited some names so that they are consistent with the convention. I can do that if you need.

bendudson:
Hey @mikekryjak Yes these are diagnostics used in the TCV-X21 energy analysis. Consistent naming would be a big improvement.

PR #332, opened 2025-04-22
recycling: Add fast momentum recycling

bendudson:
When ions fast recycle, some of their toroidal momentum is retained as the ion bounces off target, and should be transferred to the neutral. This adds fast momentum recycling factors for SOL, PF and target.


mikekryjak:
Some thoughts:
- As discussed, the parallel momentum component of the reflected neutral is non-intuitive: if the incidence angle is 0 then it's negative, if it's 45 degrees then it's 0, and if it's over 45 degrees it's positive. In this PR, the momentum recycling factor controls just the parallel component, so you can say that the angle is contained in the recycling fraction somehow.
- In reality the incidence angle is not just a function of field line and wall geometry - the particles are in a Larmor orbit which is then pulled towards the target by the sheath potential. I have heard that this ends up being a roughly similar angle a lot of the time, around 60 degrees (but the references weren't super solid on this).
- EIRENE internally calculates the ion incidence angle by taking the potential acceleration into account, and TRIM provides the momentum recycling fraction as a table for an input of material, angle and energy. I wonder whether this momentum recycling fraction is the parallel momentum specifically, allowing us to plug in 60 degrees and get an approximate value to use as default.

mikekryjak:
This definitely needs merging in in some form in the near future.

PR #333, opened 2025-04-23
Additional docs content

mikekryjak:
Added docs on:

- Form of diamagnetic drift
- Guard cells and cell boundary calculation
- Cell face notation


mikekryjak:
Need to add something on clang: 

image

bendudson (approved):

PR #355, opened 2025-06-02
Changed the order of applying the BC in Jdia

totork:
I am not sure if this does something but in FCI (from my understanding), it should be
perpendicular BC -> communication -> parallel BC

@dschwoerer


totork:
What happens to the parallel vorticity boundary condition? Do we also have to set it or is it set in the parallel sheath component? 
@dschwoerer @bendudson 

If it is not set otherwise, is it missing in the relax_potential component?

dschwoerer (approved):
Yes, the new order is the only correct order for FCI.

PR #358, opened 2025-06-06
Add a bash script to run examples

oparry-ukaea:
A convenience script that makes a copy of one of the examples in <repo_root>/runs/<example_subdir>/ and executes hermes-3 in that directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ BOUT.settings
 
 # IDE config files
 .vscode
+
+# Directory used by scripts/run_eg.sh
+runs
--- a/scripts/run_eg.sh
+++ b/scripts/run_eg.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+#
+# A convenience script to run examples.
+# Run without any arguments for usage instructions.
+
+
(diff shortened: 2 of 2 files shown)


oparry-ukaea:
Hi David,

I sort of agree with you... I created this when I first started working with hermes and didn't notice that '-d' existed.
My goal was to run an example with several different builds of the code and compare the output.
As it is, the script does make that a bit easier, but it's a long-winded way of doing it.

I guess a better way would be to allow the output directory to be set with a CL arg?

dschwoerer:
> I guess a better way would be to allow the output directory to be set 
> with a CL arg?

Ben fixed in BOUT++ that you can specify the absolute path to the input 
file (I think `-i` ), then you can just use a new directory as output 
folder with the `-d` option.

So you just need to run
`mkdir output-something && build-debug/hermes-3 -d output-something -i 
$(pwd)/examples/stellerator/transport/BOUT.inp`
or without the fix:
`mkdir myrun && build-debug/hermes-3 -d myrun -i 
../examples/stellerator/transport/BOUT.inp`

PR #359, opened 2025-06-09
Fluid equation of state with density floor

bendudson:
A relatively simple fix that seems to greatly help reduce unphysical behavior near density floors.

Defines the internal energy E as E = Cv * Nlim * T rather than the usual Cv * N * T.  This effectively puts a floor on the heat capacity of the fluid.  Then solve the pressure equation consistent with this equation of state. 

Changes made to pressure equations in `evolve_pressure`, `neutral_mixed` and `neutral_full_velocity`. Unfortunately I ran clang-format on the files so the changes seem numerous. The actual  change is quite small: In the pressure advection replace `(5/3)P` with `Nlim * T + (2/3) * P`.

This is an example from the midplane of a DIII-D simulation using `neutral_full_velocity`. The density floor is 1e-5, and previously produced quite unphysical behavior.

image

image


mikekryjak:
@bendudson am I right in thinking that this PR will have no effect if no floor is crossed?

bendudson:
Hey @mikekryjak ! I think that's right, though the `softFloor` function does change the input above the floor: It's an exponential decay so above 5 or 10 times the floor it will have negligible effect.

src/neutral_mixed.cxx
```diff
@@ -508,20 +520,23 @@ void NeutralMixed::finally(const Options& state) {
...
+    ddt(Pn) += (5. / 3)
+               * Div_a_Grad_perp_nonorthog(Dnn * e_plus_p, logPnlim, ef_adv_perp_xlow,
+                                           ef_adv_perp_ylow);
   } else {
-    ddt(Pn) +=
-        (5. / 3)
-        * Div_a_Grad_perp_flows(DnnPn, logPnlim, ef_adv_perp_xlow, ef_adv_perp_ylow);
+    ddt(Pn) += (5. / 3)
+               * Div_a_Grad_perp_flows(Dnn * e_plus_p, logPnlim, ef_adv_perp_xlow,
+                                       ef_adv_perp_ylow);
```
totork: Hi @bendudson 
Maybe I am misunderstanding something, but isn't the 5/3 already included in basically e_plus_p? That was my understanding why it is not present in the parallel advection anymore.

bendudson: Thanks, @totork ! Yes I think you are right: e_plus_p is `(5/3) * P` so the leading factor should be removed

src/evolve_pressure.cxx
```diff
@@ -418,6 +412,11 @@ void EvolvePressure::finally(const Options& state) {
       flow_ylow += get<Field3D>(species["energy_flow_ylow"]);
     }
   }
+
+  // Restore solver pressure.
+  // Keep boundary conditions for post-processing.
+  P_solver.setBoundaryTo(P);
```
mikekryjak: Why is this needed? Doesn't it introduce an inconsistency since `P_solver` uses a floored `Nn`, while `P` does not?

src/evolve_pressure.cxx
```diff
@@ -91,10 +91,6 @@ EvolvePressure::EvolvePressure(std::string name, Options& alloptions, Solver* so
   poloidal_flows =
       options["poloidal_flows"].doc("Include poloidal ExB flow").withDefault<bool>(true);
 
-  p_div_v = options["p_div_v"]
```
mikekryjak, line 94: How come we're getting rid of this?

src/evolve_pressure.cxx
```diff
@@ -278,33 +283,22 @@ void EvolvePressure::finally(const Options& state) {
...
-      //       -(4/3)*FV::Div_par(P,V) + (1/3)*(V * Grad_par(P) - P * Div_par(V))
-      //       Caused heating of charged species near sheath like p_div_v
-      ddt(P) -=
-          (5. / 3)
-          * FV::Div_par_mod<hermes::Limiter>(P, V, fastest_wave, flow_ylow_advection);
-
-      E_VgradP = V * Grad_par(P);
-      ddt(P) += (2. / 3) * E_VgradP;
-    }
     flow_ylow_advection *= 5. / 2; // Energy flow
```
mikekryjak, line 321: Since we're moving the 5/3 into the operator, it will now also be present in the flow diagnostic coming out of the operator. 

Because of this, I think the factor should be changed to 3/2 instead of 5/2.

src/neutral_mixed.cxx
mikekryjak, line 623: Should be 3/2 instead of 5/2 now that the 5/3 has gone inside the operator, see https://github.com/boutproject/hermes-3/pull/359/changes#r3577419819.

src/evolve_pressure.cxx
```diff
@@ -234,11 +230,20 @@ void EvolvePressure::transform_impl(GuardedOptions& state) {
...
+  // The equation of state is modified at low density:
+  //
+  // e = Cv T Nlim / N    <- Specific internal energy
+  // p = N T
+  //
+  // The internal energy evolution of (N * e) is therefore
+  // evolving P_solver = Nlim * T
+  // rather than pressure P = N * T
+  T = floor(P, 0.0) / softFloor(N, density_floor);
+  P_solver = P; // Save solver variable to restore later
```
mikekryjak: Could you add some comments to clearly explain the role of P, P_solver and Pint? I'm struggling to get my head around it. 

It looks like `P_solver` keeps a copy of a completely unfloored `P`, while `P` is redefined to include a `T` recalculated from `Nnlim` and hard floored to 0. Then `Pint` is `P` but with the density product also soft floored. 

So that `Pint` is the fully floored pressure while `P` is only partially floored? Is this correct? I wonder if it would be clearer if the variable names reflected the flooring state.

bendudson:
Integrated test failure is `2D-production` because the output changes slightly (1e-6, 3% max).

PR #365, opened 2025-06-16
Experimental support for heat conduction in `neutral_full_velocity`

bendudson:
Experimenting with ways to add heat conduction to the full velocity neutral model.
So far this seems to significantly slow down the model. The simple preconditioner implemented does not seem to be effective. 

Includes a modified `Laplace_FV` operator that imposes a limit on fluxes calculated at cell edges. 

Not for merging.


mikekryjak:
This will need to be picked up at some point.

PR #368, opened 2025-06-25
Add some basic tests for `sheath_boundary_simple` and refactor

ZedThree:
- Adds a golden answer regression test with five different cases, plus a unit test which has some basic sense checking
- Breaking change: make the "boundary_mode" input an enum, rather than an int. This means you can set e.g. `density_boundary_mode = limit_free` in the input file
- Bug fix: one of the loops used the wrong cell face when calculating the cross-sectional area
- Refactoring: pull out some helper functions, and use lambdas to de-duplicate the loops over the `y` boundaries


ZedThree:
> I just wanted to let you know that that is essentially what I did for sheath_parallel: https://github.com/dschwoerer/hermes-3/blob/fci-auto-with-debug/src/sheath_boundary_parallel.cxx#L226 in a way that lets you do both FCI and non-FCI in one go.

Yes, this way is very similar, although it's a complete free function, so doesn't rely on changes in BOUT++. I think a sensible plan is to get this in, and then look at refactoring it into BOUT++, and probably using regions like your version. We should very much try to avoid globals though.

> I think the plan is to only keep the parallel version, so too much refactoring is probably not worth it :-/

I'm currently merging the different sheath boundary implementations into one, so I will also pull out your parallel version and merge it all together. This PR is laying the groundwork for that.

ZedThree:
I've added a similar test for the insulating BC, along with a bug fix

mikekryjak (changes requested):
Looks good to me apart from the comment on flipping `Si_sheath` and the failed unit test.

mikekryjak:
This one is still live.

PR #372, opened 2025-07-02
Reactions regression tests

oparry-ukaea:
Adds tests to check that the results of the transforms associated with reactions don't change.

Todo:
- [x] Larger mesh ~(currently 3x5x7)~ using 8x8x8
- [x] More robust way of setting up linear input fields
- [x] Amjuel izn / rec tests
- [x] ADAS izn / rec tests
- [x] H CX tests
- [x] ADAS CX tests
- [x] Add collision_frequencies comparisons
- [x] Test ranges beyond clipped values
- ~[ ] Separate data for Release builds?~ Probably not needed
- [x] Investigate test failures on some platforms
- [x] Decide whether to relax test tolerance

### DIFF
--- a/tests/unit/test_reactions.hxx
+++ b/tests/unit/test_reactions.hxx
@@ -82,6 +82,10 @@ protected:
       axis_min = TWOPI * (mesh->zstart) / static_cast<BoutReal>(mesh->LocalNz);
       axis_max = TWOPI * (mesh->zend) / static_cast<BoutReal>(mesh->LocalNz);
       break;
+    default:
+      axis_min = axis_max = 0;
+      throw BoutException("gen_lin_field_str: no such axis type");
+      break;
     }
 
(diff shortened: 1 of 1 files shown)


oparry-ukaea:
Current test tolerances are set by allowing differences of at most [precision-of-BoutReal - N significant digits].

Some observations:

- Compiling with -O2, rather than -O0/-O1 is enough to cause some tests to fail with some compilers when N=3.
- Increasing the tolerance to N=6 is sufficient for all the tests to pass with a range of platforms, compilers and optimisation levels.

but... comparing outputs from the `tokamak-1D/1D-threshold` example:
- Builds that pass with N=3 tolerance reproduce previous diagnostic values exactly to output precision (at the same optimisation level)
- For builds that only pass with N=6 tolerance, diagnostic values differ by as much as 1e-3

PR #376, opened 2025-07-17
Update the spack config to build BOUT++ as a separate package

oparry-ukaea:
Updates the BOUT-spack submodule to make BOUT++ a separate spack package from Hermes-3
Also makes it a 'develop' package in spack.yaml.

Related to boutproject/BOUT-Spack#6 .

ToDo:
- [x] Update spack instructions in the docs
- [x] Update activate_h3env script to manage build links for BOUT
- [x] i) Delete spack-generated build links to keep submodule clean OR ii) ignore untracked in BOUT-dev OR iii) Make a PR to modify BOUT-dev/.gitignore (went for option i)
- [x] Rejig the builds links so that BOUT and hermes have consistent paths


oparry-ukaea:
N.B. Ready to review, but merge should probably wait until other BOUT-spack changes have gone in to avoid confusion.

spack.yaml
```diff
@@ -5,7 +5,7 @@
 spack:
   # add package specs to the `specs` list
   specs:
-    - hermes-3%gcc+petsc
+    - hermes-3%gcc ^boutpp+mumps+sundials
```
mikekryjak: can we have petsc mentioned in this line explicitly to make it clear to the user that it's enabled? most users/developers are not familiar with MUMPS, which is an experimental preconditioner we are testing for PETSc.

mikekryjak (changes requested):

spack.yaml
oparry-ukaea: The mumps, strumpack and superlu-dist boutpp variants have been removed, leaving those options up to petsc instead.
The default spec now installs boutpp with petsc and sundials, petsc with hypre and mumps.

mikekryjak:
Thanks for making the compile logs easy to find now. I checked the BOUT++ config summary print and found that all the settings match apart from one: `BOUT_USE_UUID_SYSTEM_GENERATOR` should be `ON` by default. 

https://github.com/boutproject/BOUT-dev/blob/ddb8855da3d124171f83e552b54f0c2096a9810b/cmake/SetupBOUTThirdParty.cmake#L349

mikekryjak (approved):

PR #377, opened 2025-07-22
Change sin_alpha to a Field3D in sheath_boundary

bshanahan:
Allows incidence angles to vary within the domain.


src/sheath_boundary.cxx
```diff
@@ -206,7 +206,7 @@ void SheathBoundary::transform(Options &state) {
                 0, 100); // Limit for e.g. Ni zero gradient
 
             // Note: Vzi = C_i * sin(α)
-            ion_sum[i] += s_i * Zi * sin_alpha * sqrt(C_i_sq);
+            ion_sum[i] += s_i * Zi * sin_alpha[ip] * sqrt(C_i_sq);
```
ZedThree, line 209: I guess the question is whether this should be `sin_alpha[ip]` or `sin_alpha[i]`. It's really only the values on the boundary that matter, so it should probably be the latter?

mikekryjak: Looks like that's a user input that is a uniform Field3D by default, so it really depends on what the user would expect in the case where it's not uniform. It would make sense to me to define the angle at the surface at the last domain cell centre, i.e. `[i]`, but it's really for @bshanahan and @dschwoerer to confirm.

If `sin_alpha` was calculated properly from an equilibrium, I think we would need to use the value at the boundary, not at the cell centre - so `(sin_alpha[i] + sin_alpha[ip])/2`

PR #381, opened 2025-08-01
fixed_velocity: Read velocity from mesh

bendudson:
If species name is "d+" then this first tries to read "Vd+0" as a Field3D from the mesh file. Should be stored in the file in SI units (m/s).

The mesh file value is used as the default value when reading from options, as before.

If neither mesh nor option value are present then an exception will be thrown.


ZedThree:
Should probably also note this in the docs!

PR #382, opened 2025-08-05
[WIP] Consolidate sheath boundaries

ZedThree:
Includes #368. ~e94eb10727065ea855cbdf388a8781b3bdfc0808 is the commit merging `sheath_boundary` into `sheath_boundary_simple`.~

~Current task is to see about merging `sheath_boundary_insulating` in as well.~

Merges `sheath_boundary`, `sheath_boundary_simple`, and `sheath_boundary_insulating` into a single implementation with compile-time switches between the different conditions.

Is `SheathKind::normal` a sensible name for the "non-simple" case? (Note that this is an internal name only, and not user-visible).

TODO:
- [ ] guard input options that aren't appropriate for the different sheath kinds


src/sheath_boundary_simple.cxx
```diff
@@ -515,173 +627,148 @@ void SheathBoundarySimple::transform(Options& state) {
...
-    }
+        // Shouldn't reach here!
+        HERMES_UNREACHABLE;
+      }();
+
+      // Negative -> into sheath
+      // TODO(peter): `normal` BC didn't allow supersonic velocities -- check this is
+      // allowed
+      const BoutReal visheath = bdry.is_lower() ? std::min(Vi[i], -sqrt(C_i_sq))
+                                                : std::max(Vi[i], sqrt(C_i_sq));
```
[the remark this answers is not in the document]
mikekryjak: Thanks for spotting this. We should always allow supersonic velocities because it helps with code stability.

mikekryjak:
`sheath_boundary` specifically allows multiple ion species, so I would maybe call it `SheathKind::multispecies`.

ZedThree:
> `sheath_boundary` specifically allows multiple ion species, so I would maybe call it `SheathKind::multispecies`.

The `simple` version also works over multiple ion species, so I'm not sure that's a useful distinguishing feature!

mikekryjak:
> > `sheath_boundary` specifically allows multiple ion species, so I would maybe call it `SheathKind::multispecies`.
> 
> The `simple` version also works over multiple ion species, so I'm not sure that's a useful distinguishing feature!

Well yes, this is technically true, but `sheath_boundary_simple` is not recommended for multiple ion species due to the lack of proper accounting in the heat transmission coefficients.

ZedThree:
> Well yes, this is technically true, but `sheath_boundary_simple` is not recommended for multiple ion species due to the lack of proper accounting in the heat transmission coefficients.

Ah, maybe it would be wise to raise a warning or something in that case? That's a separate issue I suppose

src/sheath_boundary.cxx
```diff
@@ -43,13 +59,286 @@ BoutReal limitFree(BoutReal fm, BoutReal fc) {
...
+      const BoutReal Ni_im = limitFree(Ni[ip], Ni[i], density_boundary_mode);
+      const BoutReal Ti_im = limitFree(Ti[ip], Ti[i], temperature_boundary_mode);
+      const BoutReal Te_im = limitFree(Te[ip], Te[i], temperature_boundary_mode);
+
+      // Calculate sheath values at half-way points (cell edge)
+      const BoutReal nisheath = 0.5 * (Ni_im + Ni[i]);
+      // electron temperature
+      const BoutReal tesheath = floor(0.5 * (Te_im + Te[i]), 1e-5);
+      // ion temperature
+      const BoutReal tisheath = floor(0.5 * (Ti_im + Ti[i]), 1e-5);
```
dschwoerer, line 148: Does it make sense to first interpolate in log space, and then afterwards linear?

src/sheath_boundary.cxx
```diff
@@ -43,13 +59,286 @@ BoutReal limitFree(BoutReal fm, BoutReal fc) {
...
 }
 
+/// Strong type to refer to a given boundary
+enum class YBoundarySide : bool {
+  lower, // ystart boundary
+  upper, // yend boundary
+};
+
+/// Information about a point on a Y boundary
+struct YBoundary {
```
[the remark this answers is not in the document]
dschwoerer: Can we please switch instead to [`YBoundary`](https://bout-dev--3197.org.readthedocs.build/en/3197/_breathe_autogen/file/yboundary__regions_8hxx.html#_CPPv49YBoundary) instead?

This allows to write sheath boundary code for FA and FCI!

src/sheath_boundary.cxx
dschwoerer: Yes.

That is what the code does currently.
`Ti_im` is interpolated in log-space [here](https://github.com/boutproject/hermes-3/pull/382/files#diff-ad6155cf0a77e0684d5d937deeea14b87e58a63845425d2e448be442974cf332R140)
and then in linear space, here, 7 lines later.

mikekryjak: This never occurred to me before. This feels inconsistent, I captured it in https://github.com/boutproject/hermes-3/issues/459 along with `Ni*Vi` conservation which is another possible improvement we know about.

src/sheath_boundary.cxx
```diff
@@ -43,13 +59,286 @@ BoutReal limitFree(BoutReal fm, BoutReal fc) {
...
+
+      // Calculate sheath values at half-way points (cell edge)
+      const BoutReal nisheath = 0.5 * (Ni_im + Ni[i]);
+      // electron temperature
+      const BoutReal tesheath = floor(0.5 * (Te_im + Te[i]), 1e-5);
+      // ion temperature
+      const BoutReal tisheath = floor(0.5 * (Ti_im + Ti[i]), 1e-5);
+
+      // Sound speed
+      const BoutReal C_i = sqrt((sheath_ion_polytropic * tisheath + Zi * tesheath) / Mi);
```
mikekryjak: It looks like we have different BCs in simple/insulating vs standard: the former has `Vi >= Cs` and the latter has `Vi = Cs`. The sound speed BC selection should be a user choice independent of the sheath boundary choice.

src/sheath_boundary.cxx
```diff
@@ -131,355 +451,210 @@ void SheathBoundary::transform(Options& state) {
...
+        static_assert(dependent_false<kind>, "Unhandled sheath kind");
+      }
+      HERMES_UNREACHABLE;
+      return 0.0;
+    }();
+
+    const BoutReal adiabatic = -1 - (1 / (electron_adiabatic - 1));
+
+    // Heat flux. Note: Here this is negative because vesheath < 0
+    const BoutReal _q = (_gamma_e * tesheath * nesheath * vesheath)
```
mikekryjak: Can you explain again the difference between `_q` and `q` as well as `_gamma_e` and `gamma_e`?

mikekryjak (changes requested):
In an ideal world there would be only one component called `sheath_boundary` with settings `phi_calculation` (either Tskhakhaya for multispecies or simple) and `insulating`. There would also be settings for `gamma_i`, `gamma_e` as well as the `Cs` definition (bearing in mind that `gamma_e` can only be set in `sheath_boundary_simple`). So far, this idea looks plausible but I'm not yet 100% confident that it can be done cleanly. It would be good to have another discussion and involve Ben to go through it again. I left several comments on the case switches with suggestions on how they can be replaced with flags. The biggest remaining question for me is the difference in the handling of the polytropic coefficient between `simple/insulating` and `sheath_boundary`. They assume different numbers and so would give different answers, which is unexpected. I need to read up on the formulation to figure out a form that can be used for `polytropic=1` as well as `polytropic=5/3`.

Another interesting finding is that the `Cs` assumption is sonic in `sheath_boundary` and sonic/supersonic in `sheath_boundary_simple`. This should definitely be a user setting.

Issue #348, opened 2025-05-13
Issues in pressure diagnostics


mikekryjak:
The breakdown of heat fluxes has been done incorrectly. Currently, the kinetic energy flow (e.g. `efd+_kin_ylow`) is actually only the numerical viscous heating term:

https://github.com/boutproject/hermes-3/blob/342fd2a5c5964a7e9063095c19b55035c3c62a99/src/evolve_pressure.cxx#L308

This is confusing, because you would imagine this would include the actual kinetic energy due to momentum - and if to do with viscosity, then it would include at least the kinetic energy due to Braginskii viscosity. Neither are included. 

We are also missing convective heat flux which would be really useful to have.

Suggestions:

- [ ] Rename `efd+_kin_ylow` to `efd+_kin_numerical_visc_ylow`
- [ ] Add `efd+_conv_ylow`
- [ ] Optionally add `efd+_kin_visc_ylow`

@bendudson do you remember why there would be a kinetic energy flow diagnostic for the numerical viscosity only and not for Braginskii viscosity?

mikekryjak:
Solved in https://github.com/boutproject/hermes-3/pull/385

PR #385, opened 2025-08-07
More diagnostics

mikekryjak:
Changes:
- Fix heat flow diagnostics in `evolve_pressure`. Viscous heating flux was previously labelled as kinetic. Solves https://github.com/boutproject/hermes-3/issues/348. 
- Add diagnostics for the energy and momentum transfer channels in `collisions` (energy transfer, friction heat and momentum transfer)
- Add diagnostics for the pressure gradient term in the pressure equation. At the moment the diagnostic is named after the term, which can be either `VgradP` or `PdivV`. The former is used by default and recommended for numerical reasons.
- Add intuitive descriptions to the two KE terms in ionisation and recombination (after discussions with @oparry-ukaea for #373).

Will keep this open in case I need more over the next few days.


mikekryjak:
We can merge this in now

bendudson (approved):

PR #403, opened 2025-10-03
Add term diagnostics to evolve_pressure

mikekryjak:
Adds a new flag `diagnose_terms` which saves each term in the pressure equation as a separate diagnostic. 
This is the first instance of an extra diagnostics flag in the code because it introduces a lot of new variables. Is this the right approach, or shall we have a more generic `diagnose_extra` or something like this, which would enable a generic extra level of diagnostics?

I also improved the convention for the sources diagnostic. In this PR, there is a distinct `component_source` and `user_source`. The former is a source term from other components, e.g. reactions. The latter is the user-set source. 

So, two questions:
- Should this be triggered by `diagnose_terms`, or maybe by a more generic `diagnose_extra` to act as the 2nd diagnostics level?
- Should I implement this for density and momentum equations too for consistency, or can we just start with the pressure equation for now?

todo:
- [x] Implement pressure term diagnostics in `evolve_pressure`
- [ ] Implement conduction term diagnostics in the newly separated `braginskii-conduction`
- [ ] Add docs
- [ ] From Peter: don't add more diagnose levels now. Leave that until the number of diagnostics becomes a problem, at which point we'll have more time to think about what to do.


mikekryjak:
I need to update this and merge it in.

PR #417, opened 2025-10-24
Set up repo and CI for CodeCov

cmacmackin:
This is my initial attempt to configure the CI to produce coverage reports and send them to CodeCov. Only coverage for unit tests is being pushed, although we can discuss changing that so integration test coverage is pushed too (probably with a different [flag](https://docs.codecov.com/docs/flags) to help distinguish it).

This won't work yet, as the repository still needs to have the CodeCov "app" set up and a token generated.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ defaults:
 
 jobs:
   cmake_tests:
-    name: ${{ matrix.config.name }}
+    name: ${{ matrix.config.name }}, ${{ matrix.build_type.test_type }} tests
     runs-on: ${{ matrix.config.os }}
     env:
       OMP_NUM_THREADS: ${{ matrix.config.omp_num_threads }}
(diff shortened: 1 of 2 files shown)


cmacmackin:
I'm not quite sure why the CI hasn't run for this. Possibly because no source files have been modified? Regardless, we won't actually be able to check how everything integrates with the GitHub UI for PRs until it's all been merged.

ZedThree:
> Do you think running the regression tests with a debug build would be slower than doing two separate builds? Especially considering we'd only need the debug build to be at the Hermes-3 level (BOUT++ could still be an optimised build.)

They'd run in parallel, so should be no slower. It's a fair point about BOUT++, but currently we build that as part of Hermes-3 (and don't cache it), so putting on `CMAKE_BUILD_TYPE=Debug` will also affect that

cmacmackin:
I'm not sure why the CI failure is happening. It appears that Python isn't able to find the `analysis.py` file which is in the same directory as `runtest` (`tests/integrated/drift-wave`). It all runs fine on my computer.

ZedThree (approved):

Issue #384, opened 2025-08-06
Automatically order components


cmacmackin:
Currently the `transform` (and other) methods of components get executed in the order which the components are specified in the Hermes-3 input file. This isn't ideal, as many components depend on the results of other components and this means the user must make sure to specify the desired components in the right order. This could become even more of a problem if things like collisions and other parts of the closure get put into separate components, as it would make the dependency tree that much more complicated.

I've already proposed #383, which would require components to specify which variables they read and write. Once that is implemented, it would be possible for the `ComponentScheduler` class to do construct a DAG and do a [topological sort](https://en.wikipedia.org/wiki/Topological_sorting) to determine the correct order of components. This would make Hermes-3 more reliable and easier to use.

Note: A further potential application of this is that it cold allow task-based parallelism in the execution of components. It's not clear this would be useful, given that Hermes-3 and BOUT++ already provide parallelism using domain-decomposition, however.

bendudson:
Thanks @cmacmackin ! I think this is a good way to go. One way that a `Component` could specify the options needed is to (ab)use Options again:
```
constructor(... Options& required) {
    required[name]["density"] = ReadNonBoundary;
    required[name]["temperature"] = Set;
}
``` 
or equivalently:
```
constructor(... Options& required) {
    required = {{name, 
                             {{"density", ReadNonBoundary},
                             {{"temperature", Set}};
```

The flags could be an enum class or otherwise namespaced.

I think read permissions could be:
- `Read`: Read the whole field, after which the field should not be modified
- `ReadNonBoundary` : Read the non-boundary cells, after which boundary cells may be set but non-boundary cells are locked

Write permissions may include:
- `Set` : Setting the field, which should only be done once: Species density, for example, shouldn't be set twice e.g by trying having both `evolve_density` and `fixed_density` components.
- `SetBoundary` : Set the boundary cells. Should only be done once.
- `Modify` : Modify the field. Sources, for example, are the sum of sources from reactions, equation terms etc.

Most modifications are adding terms (e.g. to sources) and so can be done in any order. More complex modifications require both reading and modifying the field e.g.:
- `electromagnetic`, which replaces all species momentum and velocity values, before boundary conditions are applied. It must be ordered after the field that sets momentum (e.g. `evolve_momentum`) but before any other components that use the velocity.
- `sheath_boundary_penalty` that both reads and modifies sources so must go after all fields that set sources e.g. reactions.

[...]

PR #421, opened 2025-11-04
Control access to state variables in transform method

### LINKED ISSUE #383: Control reading and writing of variables by components
Currently, all components can access all variables in the state. Certain components expect certain variables to have already been set, which means the user must specify the components in the correct order in the input file. It is possible for a component to overwrite some value that has already been used, which is likely to introduce bugs (there is [some functionality](https://hermes3.readthedocs.io/en/latest/developer.html#getting-setting-values) to guard against this, but its use isn't mandatory).

I propose requiring components to specify exactly from which variables they will read and to which variables they will write. They would provide this information in their constructor. The `Component` base class would provide a `transform` method which would use this information to ensure that private `transform` methods implemented by subclasses are only given access to the specified variables. This would be done with a `GuardedOptions` class that would wrap the normal `Options` class. When one attempted to access the underlying value stored in a wrapped `Options` object, it would check whether they user has permission to do so and throw an exception if not. Permission could be given for read-only or read-write.

Image

Additionally, read and write permission would be given only for specific regions of the mesh. Currently, in keeping with Hermes-3, these regions are only proposed to be "Boundary" and "Interior". In future we could potentially make these match the regions used by BOUT++. These would be represented using bit-flags, so that if one wanted a permission to apply to the entire domain then this would be represented by `Boundary | Interior`. . However,  we'd be relying on the honour system to make sure people actually only read to/write from the specified region. This is already the case for the `getNoBoundary` and `setBoundary` methods, unfortunately. I did think a bit about how we might restrict access to particular parts of the domain, but concluded this would probably take refactoring at the BOUT++ level.

cmacmackin:
This is a major refactor of how components work. They are now required to create a `Permissions` object in their constructor, which contains information on which state variables the transform method is allowed to read and write. Trying to read/write a variable without permission will result in a runtime error. This should help ensure people are more thoughtful in their design of components and prevent inadvertent side-effects. It also provides necessary information for implementing the automatic ordering of the execution of components (#384).

Closes #383

### DIFF
--- a/src/component.cxx
+++ b/src/component.cxx
@@ -66,7 +66,7 @@ bool isSetFinal(const Options& option, [[maybe_unused]] const std::string& locat
   return option.isSet();
 }
 
-bool isSetFinal(const GuardedOptions & option, const std::string& location) {
+bool isSetFinal(const GuardedOptions & option, [[maybe_unused]] const std::string& location) {
   const bool set = option.isSet();
 #if CHECKLEVEL >= 1
   const PermissionTypes perm = option.getHighestPermission();
@@ -89,7 +89,7 @@ bool isSetFinalNoBoundary(const Options& option, [[maybe_unused]] const std::str
   return option.isSet();
 }
 
-bool isSetFinalNoBoundary(const GuardedOptions & option, const std::string& location) {
+bool isSetFinalNoBoundary(const GuardedOptions & option, [[maybe_unused]] const std::string& location) {
   const bool set = option.isSet();
 #if CHECKLEVEL >= 1
   const PermissionTypes perm = option.getHighestPermission(Regions::Interior);
--- a/src/guarded_options.cxx
+++ b/src/guarded_options.cxx
@@ -72,7 +72,7 @@ void updateAccessRecords(std::map<std::string, Regions>& records, const std::str
   }
 }
 
-const Options& GuardedOptions::get(Regions region) const {
+const Options& GuardedOptions::get([[maybe_unused]] Regions region) const {
 #if CHECKLEVEL >= 1
   const std::string name = options->str();
   auto [permission, varname] = permissions->getHighestPermission(name, region);
@@ -90,7 +90,7 @@ const Options& GuardedOptions::get(Regions region) const {
 #endif
 }
 
-Options& GuardedOptions::getWritable(Regions region) {
+Options& GuardedOptions::getWritable([[maybe_unused]] Regions region) {
 #if CHECKLEVEL >= 1
   const std::string name = options->str();
   auto [access, varname] = permissions->canAccess(name, PermissionTypes::Write, region);
--- a/tests/unit/test_evolve_pressure.cxx
+++ b/tests/unit/test_evolve_pressure.cxx
@@ -63,8 +63,10 @@ TEST_F(EvolvePressureTest, Finally) {
                              {"pressure", 1.0},
                              {"temperature", 1.0},
                              {"pressure_source", 0.5}}}}}};
+#if CHECKLEVEL >= 1
   // Throws exception due to pressure_source
   EXPECT_THROW(component.finally(state), BoutException);
+#endif
 
   const Options state2 = {{"species",
                            {{"i",
--- a/tests/unit/test_guarded_options.cxx
+++ b/tests/unit/test_guarded_options.cxx
@@ -65,6 +65,7 @@ TEST_F(GuardedOptionsTests, TestGet) {
                    .isSet());
 }
 
+#if CHECKLEVEL >= 1
 TEST_F(GuardedOptionsTests, TestGetException) {
   EXPECT_THROW(guarded_opts["species:he:AA"].get(), BoutException);
   EXPECT_THROW(guarded_opts["species"]["he"]["temperature"].get(), BoutException);
@@ -78,6 +79,7 @@ TEST_F(GuardedOptionsTests, TestGetException) {
   EXPECT_THROW(guarded_opts["no_permission"].get(), BoutException);
   EXPECT_THROW(guarded_opts["species:d+:velocity"].get(), BoutException);
 }
+#endif
 
 TEST_F(GuardedOptionsTests, TestGetWritable) {
   auto& he_pressure = guarded_opts["species:he:pressure"].getWritable(Regions::Interior);
@@ -111,6 +113,7 @@ TEST_F(GuardedOptionsTests, TestGetWritable) {
   EXPECT_EQ(opts["species:d:collision_frequencies:d_t+_coll"], 14);
 }
 
+#if CHECKLEVEL >= 1
 TEST_F(GuardedOptionsTests, TestGetWritableException) {
   EXPECT_THROW(guarded_opts["species"]["he"]["temperature"].getWritable(), BoutException);
   EXPECT_THROW(guarded_opts["unset"].getWritable(), BoutException);
@@ -195,6 +198,7 @@ TEST_F(GuardedOptionsTests, TestUnwrittenItems) {
   guarded_opts["species:d:pressure"].getWritable(Regions::Interior);
   EXPECT_EQ(guarded_opts.unwrittenItems(), expected3);
 }
+#endif
 
 TEST_F(GuardedOptionsTests, TestNullOptions) {
   EXPECT_THROW(GuardedOptions(nullptr, &permissions), BoutException);
@@ -210,8 +214,10 @@ TEST_F(GuardedOptionsTests, TestGetChildren) {
   EXPECT_EQ(guarded_children.count("he"), 1);
   EXPECT_EQ(guarded_children.count("d"), 1);
   EXPECT_EQ(&(guarded_children.at("d").get()), &(opts["species"]["d"]));
+#if CHECKLEVEL >= 1
   // We do not have access to the whole "he" section
   EXPECT_THROW(guarded_children.at("he").get(), BoutException);
+#endif
 }
 
 TEST_F(GuardedOptionsTests, TestIsThisSection) {


include/hydrogen_charge_exchange.hxx
```diff
@@ -69,7 +69,7 @@ protected:
   ///  atom_energy  Energy removed from atom1, added to ion2
   ///  ion_energy   Energy removed from ion1, added to atom2
   ///
-  void calculate_rates(Options& atom1, Options& ion1, Options& atom2, Options& ion2,
+  void calculate_rates(GuardedOptions atom1, GuardedOptions ion1, GuardedOptions atom2, GuardedOptions ion2,
```
ZedThree: Another place where we're passing by value rather than by reference -- is this intentional?

include/permissions.hxx
```diff
@@ -0,0 +1,216 @@
...
+  ///
+  void substitute(const std::string& label,
+                  const std::vector<std::string>& substitutions);
+
+  /// Check whether users are allowed to access this variable to the
+  /// given permission level, in the given region. The second item
+  /// returned indicates the name of the variable or section from
+  /// which the access rights are derived. If there is no matching
+  /// section then it will be an empty string.
+  std::pair<bool, std::string> canAccess(const std::string& variable,
```
ZedThree: Is this return type basically `std::optional<std::string>`? Might be cleaner to switch to that?

include/permissions.hxx
```diff
@@ -0,0 +1,216 @@
...
+  /// string indicates the name of the variable from which the access
+  /// rights were derived. It will be empty if there are no matching
+  /// entries.
+  std::pair<std::string, AccessRights> bestMatchRights(const std::string& variable) const;
+
+  /// Return a set of access rights where the lower permissions have
+  /// been updated so that they reflect higher permissions (e.g., read
+  /// permission will be set in all cases where write permission was
+  /// set).
+  static AccessRights applyLowerPermissions(const AccessRights& rights);
```
ZedThree: Can this be a free function in the impl file?

include/permissions.hxx
```diff
@@ -0,0 +1,216 @@
...
+
+  std::map<std::string, AccessRights> variable_permissions;
+};
+
+/// Convenience function to return an object expressing that the
+/// variable should have ReadIfSet permissions on all regions.
+std::pair<std::string, Permissions::AccessRights> readIfSet(std::string varname);
```
ZedThree: Might be nicer to make the return value a `struct` with named fields?

Also, probably worth putting the definitions `inline` in the header

include/reaction.hxx
```diff
@@ -7,7 +7,7 @@
 #include "reaction_diagnostic.hxx"
 #include "reaction_parser.hxx"
 
-typedef Options& (*OPTYPE)(Options&, Field3D);
+typedef GuardedOptions (*OPTYPE)(GuardedOptions, Field3D);
```
ZedThree: If we're touching this, can we switch it to `using` rather than `typedef`?

src/permissions.cxx
```diff
@@ -0,0 +1,161 @@
...
+        result.emplace(varname, regions);
+    }
+  }
+  return result;
+}
+
+Permissions::AccessRights Permissions::applyLowerPermissions(const AccessRights& rights) {
+  AccessRights result(rights);
+  for (int i = ReadIfSet; i < PERMISSION_TYPES_END; i++) {
+    result[i] = rights[i];
```
ZedThree: Isn't this implied by the copy ctor above?

src/permissions.cxx
```diff
@@ -0,0 +1,161 @@
...
+  }
+  return result;
+}
+
+std::string Permissions::regionNames(const Regions regions) {
+  std::string result;
+  for (auto & [region, name] : fundamental_regions) {
+    if ((regions & region) == region) {
+      if (result.size() > 0) result += ", ";
+      result += name;
```
ZedThree: This looks like `fmt::join`?

ZedThree (changes requested):
I think this is broadly the right direction, although I haven't got a feel for exactly how ergonomic it is yet or how all the pieces connect exactly

include/permissions.hxx
```diff
@@ -0,0 +1,216 @@
...
+  ///
+  ///     AccessRights only_read_if_set = { AllRegions, Nowhere, Nowhere, Nowhere },
+  ///                  read_only = { Nowhere, AllRegions, Nowhere, Nowhere },
+  ///                  write_boundaries = { Nowhere, Nowhere, Boundaries, Nowhere },
+  ///                  read_and_write_everywhere = { Nowhere, AllReginos, AllRegions,
+  ///                  Nowhere
+  ///                  }, final_write_boundaries_read_interior = { Interior, Nowhere,
+  ///                      Boundaries };
+  ///
+  using AccessRights = std::array<Regions, PERMISSION_TYPES_END>;
```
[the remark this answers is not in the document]
cmacmackin: I suppose that would be nicer, although in reality people will use the [helper methods](https://github.com/boutproject/hermes-3/pull/421/files#diff-75fec155c89e141183a33643cfb09df3743783ca46bb581062abd32693291a4cR202) in at least 95% of cases.

src/permissions.cxx
cmacmackin: Yes, I guess it is. I wasn't aware of that function.

src/guarded_options.cxx
```diff
@@ -0,0 +1,113 @@
...
+    // Only add variables with permission ReadIfSet to
+    // unread_variables if they are already present in the options
+    // object
+    if (options != nullptr) {
+      for (auto& [varname, region] :
+           permissions->getVariablesWithPermission(Permissions::ReadIfSet)) {
+        // find the last component of the full varname path
+        size_t last_colon = varname.find_last_of(":");
+        Options& parent = (last_colon != std::string::npos)
+                              ? (*options)[varname.substr(0, last_colon)]
```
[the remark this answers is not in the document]
cmacmackin: I think your suggestion here is wrong. Shouldn't it be `(*options)[varname].parent()`? The reason why I didn't do that (or, better yet, just simplify all of this logic to `(*options)[varname].isSet()`) is I wanted to avoid creating Option nodes that don't already exist.

include/component.hxx
```diff
@@ -96,6 +111,10 @@ T getNonFinal(const Options& option) {
                         option.str(), typeid(T).name());
   }
 }
+template<typename T>
+T getNonFinal(const GuardedOptions option) {
```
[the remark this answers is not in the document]
cmacmackin: OK, I've read up on this a bit. In places where GuardedOptions is const I've switched to passing by reference.

src/permissions.cxx
cmacmackin: Yes, you're right.

include/permissions.hxx
cmacmackin: You are right that this would express the same information. I'm not sure whether that's the best approach though. The main information being returned here is whether the access rights exist. The variable name from which those rights are derived is secondary information. I'm not sure it would make sense to effectively make that the primary information by returning a `std::optional<std::string>`.

src/guarded_options.cxx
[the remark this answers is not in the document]
cmacmackin: I've since refactored this anyway, so we avoid creating sections that don't already exist. Otherwise we end up getting bugs where empty sections have been created and are assumed by components to contain certain variables.

include/permissions.hxx
cmacmackin: Done.

include/component.hxx
```diff
@@ -10,19 +10,76 @@
...
+                     const std::vector<std::string>& positive_ions,
+                     const std::vector<std::string> & negative_ions)
+    : electrons(electrons), neutrals(neutrals), positive_ions(positive_ions), negative_ions(negative_ions), ions(positive_ions) {
+    finish_construction();
+  }
+
+  SpeciesInformation(const std::initializer_list<std::string> species) {
+    for (auto& sp : species) {
+      // FIXME: identifySpecies only identifies positive ions
+      // FIXME: identifySpecies has no concept of ebeam
```
ZedThree: Should we make these issues to track them?

include/component.hxx
```diff
@@ -47,6 +104,37 @@ struct Component {
                                            const std::string &name, // The species/name for this instance
                                            Options &options,  // Component settings: options[name] are specific to this component
                                            Solver *solver); // Time integration solver
+
+  /// Tell the component the name of all species in the simulation, by type. It
+  /// will use this information to substitute the following placeholders in `svate_variable_access`:
```
ZedThree:
```suggestion
  /// will use this information to substitute the following placeholders in `state_variable_access`:
```

src/classical_diffusion.cxx
```diff
@@ -2,28 +2,47 @@
...
   Bsq = SQ(bout::globals::mesh->getCoordinates()->Bxy);
 
   diagnose = options["diagnose"].doc("Output additional diagnostics?").withDefault<bool>(false);
   custom_D = options["custom_D"].doc("Custom diffusion coefficient override. -1: Off, calculate D normally").withDefault<BoutReal>(-1);
+
+  state_variable_access.substitute(
+      "optional", {"charge", "pressure", "density", "velocity", "temperature"});
+  std::vector<std::string> e_vals = {"AA", "density"};
+  if (custom_D <= 0.)
+    e_vals.push_back("collision_frequency");
```
ZedThree:
```suggestion
  if (custom_D <= 0.) {
    e_vals.push_back("collision_frequency");
  }
```

src/component.cxx
```diff
@@ -24,10 +58,39 @@ bool isSetFinal(const Options& option, const std::string& location) {
...
   const_cast<Options&>(option).attributes["final-domain"] = location;
 #endif
   return option.isSet();
 }
+
+bool isSetFinalNoBoundary(const GuardedOptions & option, const std::string& location) {
+  bool set = option.isSet();
+#if CHECKLEVEL >= 1
+  PermissionTypes perm = option.getHighestPermission(Regions::Interior);
+  if (static_cast<int>(perm) >= static_cast<int>(PermissionTypes::Read)
```
ZedThree: No need to cast

src/component.cxx
```diff
@@ -24,10 +58,39 @@ bool isSetFinal(const Options& option, const std::string& location) {
...
 }
 
+bool isSetFinal(const GuardedOptions & option, const std::string& location) {
+  bool set = option.isSet();
+#if CHECKLEVEL >= 1
+  PermissionTypes perm = option.getHighestPermission();
+  if (static_cast<int>(perm) >= static_cast<int>(PermissionTypes::Read)
+      or (perm == PermissionTypes::ReadIfSet and set)) {
+    const Options& opt = option.get();
+    const_cast<Options&>(opt).attributes["final"] = location;
```
ZedThree: I really don't like this. It doesn't look like `isSetFinal` is called in any context with a `const&` anyway? I know this is already an issue with the existing function, but can we avoid the `const_cast` here?

src/guarded_options.cxx
```diff
@@ -0,0 +1,113 @@
...
+    : options(options), permissions(permissions),
+      unread_variables(std::make_shared<std::map<std::string, Permissions::Regions>>()),
+      unwritten_variables(
+          std::make_shared<std::map<std::string, Permissions::Regions>>()) {
+  if (permissions != nullptr) {
+    *unread_variables = permissions->getVariablesWithPermission(Permissions::Read);
+    // Only add variables with permission ReadIfSet to
+    // unread_variables if they are already present in the options
+    // object
+    if (options != nullptr) {
```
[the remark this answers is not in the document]
ZedThree: Is it possible to delete the default ctor?

src/permissions.cxx
```diff
@@ -0,0 +1,174 @@
...
+  return result;
+}
+
+void Permissions::printAllPermissions() const {
+  for (const auto& [varname, perms] : variable_permissions) {
+    fmt::print("{}", varname);
+    for (int i = 0; i < static_cast<int>(PermissionTypes::END); i++) {
+      fmt::print(" {}", static_cast<int>(perms[i]));
+    }
+    fmt::print("\n");
```
ZedThree: Is this just:
```suggestion
    fmt::print("{} {}\n", varname, fmt::join(" ", perms));
```

Alternatively, replace this function with an fmt formatter, then it's possible to convert to a string and use in more contexts

include/permissions.hxx
```diff
@@ -0,0 +1,277 @@
...
+  ///     AccessRights only_read_if_set = { Regions::All, Regions::Nowhere,
+  ///                      Regions::Nowhere, Regions::Nowhere },
+  ///                  read_only = { Regions::Nowhere, Regions::All, Regions::Nowhere,
+  ///                      Regions::Nowhere },
+  ///                  write_boundaries = { Regions::Nowhere, Regions::Nowhere,
+  ///                      Regions::Boundaries, Regions::Nowhere },
+  ///                  read_and_write_everywhere = { Regions::Nowhere,
+  ///                      Regions::AllReginos, Regions::All, Regions::Nowhere },
+  ///                  final_write_boundaries_read_interior = { Regions::Nowhere,
+  ///                      Regions::Interior, Regions::Nowhere, Regions::Boundaries };
```
ZedThree: Please put each declaration as a separate statement, it vastly improves the readability, which is even more important for the docs!

ZedThree (changes requested):
I've made a bunch of mostly minor comments. I realised why I've been struggling to get my head around some of the implementation details, and that's because the design feels a little backwards. We have an array where each index is a permission level and the element is a region. But if we look at say unix permissions, we have a structure where the position is the user/group and the value is the permission level (and this is a bit-field, being able to set `read | write` permissions).

What if we had something like this:

```cpp
enum class AccessRights { none, read_if_set, read, write, final_ };

enum class PermissionRegions { interior, boundaries };

struct AccessPermission {
  AccessRights interior = AccessRights::none;
  AccessRights boundaries = AccessRights::none;

  auto operator[](PermissionRegions region) {
    switch(region) {
    case PermissionRegions::interior:
       return interior;
    case PermissionRegions::boundaries:
       return boundaries;
    }
  }
};
```

Is this way round easier to reason about? It feels a lot simpler. To take the examples from the `AccessRights` docstring, we'd now have:

```cpp
auto only_read_if_set = AccessPermission {
  .interior = AccessRights::read_if_set,
  .boundaries = AccessRights::read_if_set
};

auto read_only = AccessPermission {
  .interior = AccessRights::read,
  .boundaries = AccessRights::read
};

auto write_boundaries = AccessPermission {
  .interior = AccessRights::none,
  .boundaries = AccessRights::write
};

auto read_write_everywhere = AccessPermission {
  .interior = AccessRights::write,
  .boundaries = AccessRights::write
};

auto final_write_boundaries_read_interior = AccessPermission {
  .interior = AccessRights::read,
  .boundaries = AccessRights::write
};
```

Checking if a `AccessPermission` is valid for a given region and access level is just:

```cpp
read_only[region] >= AccessRights::read
```

This way round would of course also work with an array, but I keep suggesting structs because names are _really_ important, and the lack of them in the current design makes e.g. `applyLowerPermissions` very difficult to parse. I'm not even sure it's needed this proposed design.

Applying the higher of two permissions is easier:

```cpp
auto result = AccessPermission {
  .interior = std::max(lhs.interior, rhs.interior),
  .boundaries = std::max(lhs.boundaries, rhs.boundaries),
};
```

cmacmackin:
The other problem with refactoring the way permissions are stored along the lines @ZedThree suggests is that I'm only on this project for another week, so it's not clear I'd have time to do that, get it reviewed, etc.

src/permissions.cxx
```diff
@@ -0,0 +1,244 @@
...
+    const auto [varname, access] = *it;
+    const std::string pattern = "{" + label + "}";
+    if (varname.find(pattern) == std::string::npos) {
+      it++;
+      continue;
+    }
+    it = variable_permissions.erase(it);
+    for (const std::string& val : substitutions) {
+      const std::string newname = replaceAll(varname, pattern, val);
+      // Do not overwrite permissiosn that are already set
```
mikekryjak:
```suggestion
      // Do not overwrite permissions that are already set
```

src/vorticity.cxx
```diff
@@ -205,13 +206,42 @@ Vorticity::Vorticity(std::string name, Options& alloptions, Solver* solver) {
   diagnose = options["diagnose"]
     .doc("Output additional diagnostics?")
     .withDefault<bool>(false);
+
+  if (diamagnetic or diamagnetic_polarisation) {
```
mikekryjak: Is it actually necessary to change the permissions on such a granular level, and has this been done to a consistent level of granularity across the components?

cmacmackin:
@ZedThree I've created issues for all of your comments I have not implemented. Could you mark this as approved (or at least remove the request for changes)?

include/component.hxx
```diff
@@ -47,6 +115,58 @@ struct Component {
                                            const std::string &name, // The species/name for this instance
                                            Options &options,  // Component settings: options[name] are specific to this component
                                            Solver *solver); // Time integration solver
+
+  /// Tell the component the name of all species in the simulation, by type. It
+  /// will use this information to substitute the following placeholders in
+  /// `state_variable_access`:
+  ///   - electrons (any electron species)
+  ///   - electrons2 (same as above, used for cross-product)
```
mikekryjak: I've not been able to figure out what you mean by cross-product here - can you explain?

Also, I think this list should be in the docs. Maybe they should be pulled out of the developer manual and put into its own top level category along with the rest of `developer manual/code structure`.

include/component.hxx
```diff
@@ -47,6 +115,58 @@ struct Component {
...
+
+private:
+  /// Information on which state variables the transform method will read and write.
+  Permissions state_variable_access;
+
+  /// Modify the given simulation state. All components must
+  /// implement this function. It will only allow the reading
+  /// from/writing to state variables with the appropriate permissiosn
+  /// in `state_variable_access`.
+  virtual void transform_impl(GuardedOptions &state) = 0;
```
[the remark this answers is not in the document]
cmacmackin: `impl` stands for "implementation`. I'd originally planned just to make it an overload of `transform`, but there seemed to be some confusing rule about overload resolution that was causing me problems, so in the end I gave it a different name.

src/evolve_density.cxx
```diff
@@ -15,7 +15,7 @@
 using bout::globals::mesh;
 
 EvolveDensity::EvolveDensity(std::string name, Options& alloptions, Solver* solver)
-    : name(name) {
+    : Component({readWrite("species:{name}:{outputs}")}), name(name) {
```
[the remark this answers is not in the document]
cmacmackin: This doesn't actually seem to happen in `evolve_energy` though.

include/component.hxx
cmacmackin: The list is included in the docs as part of the class documentation for `Component`. I do also refer to them else where. I'm a little reluctant to write them out more than once, as that's just asking for the two lists to get out of sync. Do you think there is somewhere else that would be better?

src/sheath_boundary_simple.cxx
```diff
@@ -57,8 +57,25 @@ BoutReal limitFree(BoutReal fm, BoutReal fc, BoutReal mode) {
...
-                                           Solver*) {
+SheathBoundarySimple::SheathBoundarySimple(std::string name, Options& alloptions, Solver*)
+    : Component({
+        readIfSet("species:e:{e_whole_domain}"),
+        writeBoundary("species:e:{e_boundary}"),
+        readWrite("species:e:energy_source"),
+        readWrite("species:e:energy_flow_ylow"),
+        writeBoundaryIfSet("species:e:{e_optional}"),
+        {"species:e:pressure",
+         {Regions::Interior, Regions::Nowhere, Regions::Boundaries, Regions::Nowhere}},
```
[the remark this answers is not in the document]
cmacmackin: This means to read the pressure in the interior if it has been set, but to set the pressure on the boundaries whether it's been set or not.

src/sheath_boundary_simple.cxx
```diff
@@ -57,8 +57,25 @@ BoutReal limitFree(BoutReal fm, BoutReal fc, BoutReal mode) {
...
+        // FIXME: These only applies to ions, not to all species
```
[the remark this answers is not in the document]
cmacmackin: This is something which I should have fixed already but overlooked. the `{all_species}` should have been changed to `{ions}`. I've done that now.

docs/sphinx/developer.rst
```diff
@@ -516,21 +521,211 @@ Notes:
...
+Write
+   Can write variable. Makes no assumption about whether it has already
+   been written or will be written again in future.
+
+Final
+   This will be the last component to write to the variable. Only one
+   component may have ``Final`` permission for a given variable.
+
+The order these per permissions are listed in is significant: each
+permission implies a component also has all lower permissions. E.g.,
```
mikekryjak: Need to define "lower permissions" and the hierarchy here.

ZedThree (approved):

PR #422, opened 2025-11-05
[WIP] Add `freeze_profiles` option to fix the Z-average profiles

bendudson:
Removes the DC, i.e. Z-averaged, component of the time-derivatives. This effectively adds artificial, time-varying, sources and sinks to keep the profiles fixed. Not recommended for nonlinear simulations, but useful for linear analysis.


ZedThree:
Is this something that needs to be turned on consistently across components? If so, would it be better as a component itself, and just set something in the state?

bendudson:
@ZedThree There may be cases where some species should be frozen and not others e.g. freeze plasma profiles and evolve neutrals or vice-versa. I'm not against adding a component that sets a flag that is picked up by all components.

PR #423, opened 2025-11-07
FCI Tests and options for testing

totork:
This PR is aiming to add the following things:

- 1D FCI tests for parallel flow, heat conduction, ion/electron viscosity. Amid these tests, changes for the Div_par_mod operator are introduced that allows the use of a dissipative form. If this is used, the 2nd order convergence is not achieved. 
- 2D FCI tests for the perpendicular diffusion of density and pressure. During the tests, it appeared that the dagp operator is not properly working for slab grids and another operator is introduced (even though intermediately) to cross check. 
**- Both tests have their required files are tar files added.**
- For these tests, I introduced some variables that allows me to disable the time evolution or the double adding of source terms

~~A crucial physics change is the use of **only** the self-collision time for each species conduction/viscosity. This should be in line with Braginskii? The proper treatment shall be pulled from master, but it is getting refactored now~~
- I began working on the relax potential component. 
- Added option to not extrapolate into sheath. I had mixed experience with extrapolation in FCI, so it might be worth testing if simulations are unstable

### DIFF
--- a/src/evolve_pressure.cxx
+++ b/src/evolve_pressure.cxx
@@ -218,6 +218,7 @@ void EvolvePressure::transform(Options& state) {
   }
 
   mesh->communicate(P);
+  P.applyParallelBoundary();
 
   if (neumann_boundary_average_z) {
     // Take Z (usually toroidal) average and apply as X (radial) boundary condition
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,7 +60,7 @@ jobs:
           echo ~/.local/bin >> $GITHUB_PATH
       - name: Build and run tests
         run: cmake . -B build $BOUT_CONFIGURE_OPTIONS &&
-             cmake --build build &&
+             cmake --build build -j 2 &&
              cd build &&
              ctest --output-on-failure --timeout 300
         env:
--- a/.pip_install_for_tests.sh
+++ b/.pip_install_for_tests.sh
@@ -3,8 +3,8 @@
 set -e
 
 export PATH=${HOME}/.local/bin:${PATH}
-pip3 install --user --upgrade pip~=20.0 setuptools~=46.1
-pip3 install --user --upgrade scipy~=1.4 numpy~=1.18
+pip3 install --user --upgrade pip setuptools
+pip3 install --user --upgrade scipy numpy
 for package in $@
 do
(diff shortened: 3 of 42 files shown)


bendudson:
Thanks @totork ! 

I'm not sure about the self-collisions thing: Scattering from any species will slow diffusion. Fitzpatrick's notes on Braginskii use electron-ion collision frequency (https://farside.ph.utexas.edu/teaching/plasma/Plasma/node53.html). I think @mikekryjak discussed this with Stefan so maybe has suggestions.

totork:
I think it might be wise to keep it like this now. It is already quite alot of changes. Changed to be ready @bendudson @dschwoerer

src/ion_viscosity.cxx
```diff
@@ -92,6 +92,7 @@ void IonViscosity::transform(Options &state) {
       continue;
     }
 
+    //const Field3D tau = 1. / get<Field3D>(species["collision_frequency"]);
```
bendudson:
```suggestion
```

src/evolve_pressure.cxx
```diff
@@ -364,7 +377,9 @@ void EvolvePressure::finally(const Options& state) {
   if (thermal_conduction) {
 
     // Calculate ion collision times
-    const Field3D tau = 1. / floor(get<Field3D>(species["collision_frequency"]), 1e-10);
+    //const Field3D tau = 1. / floor(get<Field3D>(species["collision_frequency"]), 1e-10);
```
bendudson:
```suggestion
```

src/evolve_momentum.cxx
```diff
@@ -211,8 +224,13 @@ void EvolveMomentum::finally(const Options &state) {
   //  - Density floor should be consistent with calculation of V
   //    otherwise energy conservation is affected
   //  - using the same operator as in density and pressure equations doesn't work
-  ddt(NV) -= AA * FV::Div_par_fvv<hermes::Limiter>(Nlim, V, fastest_wave, fix_momentum_boundary_flux);
-  
+
+  if (isMMS) {
```
bendudson: This feels a bit wrong, to use a different method in MMS testing than in production. The `Div_par_fvv` operator converges for non-FCI, so the issue must be in the FCI branch of `Div_par_fvv`.

PR #428, opened 2025-11-27
Perform topological sort to order components

### LINKED ISSUE #384: Automatically order components

cmacmackin:
Using the access control information introduced in #421, this PR makes it possible for Hermes-3 to work out the order of components at run-time. This will make things far simpler and more robust for users. It will also fail faster if there is an unsatisfiable or circular dependency.

Closes #384.


cmacmackin:
This is nearly done. Remaining tasks:

- [ ] Merge #421 and rebase.
- [ ] There is one integration test giving me strange h5py errors on my computer. It doesn't look to be related to any changes I've made. We'll see what happens in CI. I seem to recall we saw something similar elsewhere at one point.
- [ ] Update documentation.

src/component_scheduler.cxx
```diff
@@ -1,13 +1,314 @@
...
+#include <fmt/format.h>
+#include <fmt/ranges.h>
 
 #include "../include/component.hxx"
 #include "../include/component_scheduler.hxx"
+#include "../include/permissions.hxx"
+
+const std::set<std::string> ComponentScheduler::predeclared_variables = {
+    "time",          "linear",      "units:inv_meters_cubed", "units:eV", "units:Tesla",
+    "units:seconds", "units:meters"};
```
ZedThree: Given this is only used in this file, we can move the declaration out of the header and just have it here.

docs/sphinx/developer.rst
```diff
@@ -662,8 +679,10 @@ of components
    # options to control component3
 
 This will create three components, which will be run in the order
```
mikekryjak: I think this is incorrect, because the run order will now be determined by the component scheduler.

include/permissions.hxx
```diff
@@ -308,6 +317,20 @@ inline Permissions::VarRights writeBoundary(std::string varname) {
...
+  return {std::move(varname),
+          {Regions::Interior, Regions::Nowhere, Regions::Boundaries, Regions::Nowhere}};
+}
+
+/// Convenience function to return an object expressing that, if the
+/// interior has been set, then the variable should have Final write
+/// permissions on the boundaries and read permissions for the
+/// interior.
+///
+/// FIXME: Currently these permissiosn are not expressed properly, due
```
mikekryjak, line 328: is this FIXME still active? what does it take to fix it and what are the possible risks of leaving it unfixed?

src/component_scheduler.cxx
```diff
@@ -1,13 +1,314 @@
...
+  processing[item] = true;
+
+  for (const auto dep : dependencies[item]) {
+    topological_sort(dependencies, dep, sorted, processing, processed);
+  }
+  processed[item] = true;
+  sorted.push_back(item);
+}
+
+/// Get all the parent sections of a variable "path". Sections are
```
mikekryjak, line 134: Can you clarify that the path is just the hierarchy of the state object.

src/component_scheduler.cxx
```diff
@@ -1,13 +1,314 @@
...
+    }
+  }
+
+  /// Assemble the set of all section names which are referred to
+  /// explicitly in the component permissions.
+  std::set<std::string> sections_present;
+  std::set_intersection(unconditional_names.begin(), unconditional_names.end(),
+                        unconditional_sections.begin(), unconditional_sections.end(),
+                        std::inserter(sections_present, sections_present.begin()));
+  /// Assemble the set of all variable names which are definitlye
```
mikekryjak: Typo

mikekryjak:
I missed this in #421, but I see that you added a long-needed feature - a better way to determine species type which is based on the charge:
https://github.com/boutproject/hermes-3/blob/8fc17e0cad037999b97a31cdf53d6558cdd859d8/src/component_scheduler.cxx#L339-L352

There is already a function that does the same in `hermes_utils` and is used in a few places in the code:
https://github.com/boutproject/hermes-3/blob/053c9bb70b66ffd3ac603477f76c5866a0b9e33c/include/hermes_utils.hxx#L61-L70

Ideally there should be only one tool to do this, and your new one seems more robust. Is there any reason not to replace the one in `hermes_utils` with the new one?

cmacmackin:
There are some places (e.g., when writing tests) where it was convenient to just be able to have a list of species names and use the old heuristics to categorise them. Probably not a good enough reason to keep that though.

ZedThree (approved):
LGTM, thanks @cmacmackin!

There's some trivial bits that I'm happy to fix myself

src/component_scheduler.cxx
```diff
@@ -1,13 +1,314 @@
...
+/// Perform a depth-first topological sort, starting from `item`. It
+/// will finish once it reaches the end of `item`'s dependency chain,
+/// so this needs to be called in a loop for all items. Information
+/// about `item` is stored in the corresponding index of the vector
+/// arguments.
+///
+/// In pratice, `item` here represents the index of a particular
+/// component. The indices of the dependencies of `item` are stored in
+/// the corresponding element of `dependencies`.
+void topological_sort(const std::vector<std::set<size_t>>& dependencies, size_t item,
```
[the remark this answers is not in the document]
ZedThree: It's a guard against One Definition Rule (ODR) violations -- basically ensures that functions have names unique to the translation unit, and so if there happens to be another function with the same name somewhere else, they won't clash at link time.

cmacmackin:
I've merged in the changes from #596 as they are useful to writing error messages during the topological sort process.

docs/sphinx/developer.rst
```diff
@@ -577,6 +577,17 @@ The `name` is a string labelling the instance. The `alloptions` tree contains at
...
+All component constructors must pass a `Permissions` object to the
+constructor on the `Component::Component` base class. This specifies
+which variables will be read/written by the `Component::transform`
+method and will be used to construct a `GuardedOptions` object to be
+passed into `Component::transform_impl`. The `Permissions` object
+(`Component::state_variable_access`) can be further updated in the
+body of the constructor of your component, based on what
+configurations were specified in ``alloptions``. You should give read
+and write permissions to the minimum number of variables necessary, to
+avoid circular dependencies arising among components.
```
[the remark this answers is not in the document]
cmacmackin: Fixed

src/component_scheduler.cxx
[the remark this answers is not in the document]
cmacmackin: Done this now.

src/component_scheduler.cxx
```diff
@@ -1,13 +1,314 @@
...
+
+  std::map<std::string, std::set<std::string>> result;
+
+  // ReadIfSet variables will only actually be used if they are
+  // reference elsewhere. We create them with empty sets, which will
+  // get filled if they are present.
+  for (const auto& name : conditional_names) {
+    // FIXME: this isn't an empty set
+    //result.insert({name, {name}});
+    result.insert({name, {}});
```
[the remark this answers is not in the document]
cmacmackin: Deleted this

src/component_scheduler.cxx
```diff
@@ -1,13 +1,314 @@
...
+  for (const auto dep : dependencies[item]) {
+    topological_sort(dependencies, dep, sorted, processing, processed);
+  }
+  processed[item] = true;
+  sorted.push_back(item);
+}
+
+/// Get all the parent sections of a variable "path". Sections are
+/// separated by colons in the path.
+std::set<std::string> getParents(const std::string& name) {
```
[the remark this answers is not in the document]
cmacmackin: I've expanded the docstring to explain this better.

src/component_scheduler.cxx
cmacmackin: Done.

bendudson:
I think this should be merged but with an input option (`hermes:topological_sort`?) that is off by default (for now) to control whether `::sortComponents(components)` is called in `component_scheduler.cxx`. That would (a) simplify testing on real cases, reducing the risks of breaking something; (b) prevent accumulation of merge conflicts if this remains unmerged. I think sorting should be enabled in the integrated tests.

cmacmackin:
The formatter was added in #596 and then merged into this branch. I wanted #596 to be merged before the final review of this PR. It would be better if you pushed [d255141](https://github.com/boutproject/hermes-3/pull/428/commits/d255141508b4f26e97189350f810fe320efe4c16) there instead. Similarly, I've rebased #596 off of master and updated the commit history to deal with `external_apar`. That replicated commits [833d87c](https://github.com/boutproject/hermes-3/pull/428/commits/833d87c41121fbad3a91ae32ea40cb74618aed18) to [fc62ffe](https://github.com/boutproject/hermes-3/pull/428/commits/fc62ffe2dcf2ff59f46eb390df203a0be337749e). Once That PR was merged into master I intended to update this branch accordingly.

mikekryjak:
We should remove mention of `ebeam` - electrons should only be named as `e`

bendudson:
> We should remove mention of `ebeam` - electrons should only be named as `e`

Yes this should probably be removed. It was for modeling of LAPD, where it's useful to have two species of electrons: A high energy beam (`ebeam`) and the usual low energy electrons (`e`). It's probably not a common use case, and e.g. reaction rates should be calculated quite differently for a beam than a Maxwellian.

cmacmackin:
There are a few examples that failed when I applied the topological sort to their components. 

In `other/linear/annulus-isothermal-he-emag`, both `electromagnetic` and `sheath_boundary` attempt to write the boundaries of the momentum and velocity of electrons and helium ions. This is something both components legitimately do. I'm guessing that `sheath_boundary` is intended to overwrite the boundary values set by `electromagnetic`. I've updated the latter accordingly so it no longer sets boundary values for these variables.

`tokamak-1D/extra/1D-hydrogen-equalT` had both `sheath_boundary` and `transform` trying to write the boundaries of deuterium ion temperature. I updated the permissions of `transform` so it is only considered to read and write the interior of a variable, not the boundaries. This could hypothetically cause problems in the future if someone expects it to write the boundaries. I could potentially make it configurable whether it write the boundaries or not.

`tokamak-2D/heat-transport` had a small mistake which meant it was asking for permission to read temperatures in the boundaries, when it didn't need to. This mistake resulted in an runtime error due to a circular dependency between components. The circular dependency went away once the permissions were corrected.

I didn't have time to wait for all the tests to finish before I left for the day, so we'll have to see whether these changes introduced any regressions.

src/component.cxx
```diff
@@ -102,6 +102,29 @@ bool isSetFinalNoBoundary(const GuardedOptions& option,
...
+
+bool isSetFinalBoundary(const GuardedOptions& option,
+                        [[maybe_unused]] const std::string& location) {
+  const bool set = option.isSet();
+#if CHECKLEVEL >= 1
+  const PermissionTypes perm = option.getHighestPermission(Regions::Boundaries);
+  if (perm >= PermissionTypes::Read or (perm == PermissionTypes::ReadIfSet and set)) {
+    // Mark option as final inside the domain, but not in the boundary
+    const_cast<Options&>(option.get(Regions::Interior)).attributes["final-bounds"] =
+        location;
```
bendudson: This comment and assignment are probably wrong: I think this should be setting the boundary flag not the interior.
```suggestion
    // Mark option as final in the boundaries but not the interior
    const_cast<Options&>(option.get(Regions::Boundaries)).attributes["final-bounds"] =
        location;
```

tests/unit/test_component.cxx
```diff
@@ -138,6 +162,22 @@ TEST(ComponentTest, IsSetFinalStaysFalse) {
...
+}
+
+TEST(ComponentTest, IsSetFinalBoundaryStaysFalse) {
+  Options option;
+
+  ASSERT_EQ(isSetFinalBoundary(option["test"]), false);
+  // Shouldn't change if called again
+  ASSERT_EQ(isSetFinalBoundary(option["test"]), false);
+}
+
```
bendudson, line 187: GPT suggests a test to cover the `GuardedOptions` overload of `IsSetFinalBoundary`
```suggestion
/// isSetFinalBoundary should succeed without needing interior access, 
/// lock boundary writes, and still allow interior writes.
TEST(ComponentTest, IsSetFinalBoundaryGuardedBoundaryOnly) {
  Options options;
  options["test"] = 1;

  Permissions permissions({readOnly("test", Regions::Boundaries)});
  GuardedOptions guarded(&options, &permissions);

  ASSERT_TRUE(isSetFinalBoundary(guarded["test"]));

#if CHECKLEVEL >= 1
  // Boundary is now assumed final
  ASSERT_THROW(setBoundary<int>(options["test"], 2), BoutException);

  // Interior/domain is still allowed
  ASSERT_NO_THROW(setNoBoundary<int>(options["test"], 3));
  ASSERT_EQ(getNoBoundary<int>(options["test"]), 3);
#endif
}
```

docs/sphinx/developer.rst
```diff
@@ -314,14 +315,15 @@ the domain cells, leaving the guard cells to be settable using ``getNoBoundary``
...
 
 .. code-block:: ini
 
    Options& setBoundary(Options& option, T value) {
    // Check that the value has not already been used
    #if CHECKLEVEL >= 1
-   if (option.hasAttribute("final")) {
+   if (option.hasAttribute("final-bounds")) {
       throw BoutException("Setting boundary of {} but it has already been used in {}.",
                            option.name(), option.attributes["final"].as<std::string>());
```
bendudson: "final" was changed to "final-bounds" in the conditional but not in the throw.
```suggestion
                           option.name(), option.attributes["final-bounds"].as<std::string>());
```

docs/sphinx/developer.rst
```diff
@@ -286,7 +286,7 @@ already been "gotten", they can't be set again:
    Options& set(Options& option, T value) {
    // Check that the value has not already been used
    #if CHECKLEVEL >= 1
-   if (option.hasAttribute("final")) {
+   if (option.hasAttribute("final-bounds")) {
       throw BoutException("Setting value of {} but it has already been used in {}.",
                            option.name(), option.attributes["final"].as<std::string>());
```
bendudson:
```suggestion
                           option.name(), option.attributes["final-bounds"].as<std::string>());
```

docs/sphinx/developer.rst
```diff
@@ -301,8 +301,9 @@ already been "gotten", they can't be set again:
    }
    #endif
 
-There is a special use case which allows you to use this "locking" scheme for only
-the domain cells, leaving the guard cells to be settable using ``getNoBoundary``:
+There are special use cases which allows you to use this "locking" scheme for only
+the domain (guard) cells, leaving the guard (domain) cells
+to be settable using ``getNoBoundary`` (``getBoundary``) :
```
bendudson: This use of parentheses makes my brain ache. 

```suggestion
There are special use cases which allow you to use this "locking" scheme for only some cells: ``getNoBoundary`` locks the domain cells while leaving the boundary cells settable. ``getBoundary`` locks the boundary cells while leaving the domain cells settable.
```

docs/sphinx/developer.rst
```diff
@@ -670,12 +678,49 @@ of components
    # options to control component3
 
 This will create three components, which will be run in the order
-`component1`, `component2`, `component3`: First all the components
-in `group1`, and then `component3`.
+`component1`, `component2`, `component3`: First all the components in
+`group1`, and then `component3`. Grouped components will be sorted
+individually when determining the run order; they may not be run
+together.
```
bendudson: The ordering isn't preserved and all components are flattened into a single list before sorting. 
```suggestion
determined by the topological sorting algorithm. First all the components in
`group1`, and then `component3` are flattened into a single list. The access permissions of each component, i.e. which quantities each component reads or writes, are used to put the components in order. This means that grouped components may not run together: The scheduler may run `component1` then `component3` then `component2` if the inputs and outputs of the components require this.
```

include/permissions.hxx
bendudson: Does the section name still need to end in a colon?

docs/sphinx/inputs.rst
```diff
@@ -13,14 +13,15 @@ format is described in the `BOUT++ options manual
...
+The first section contains top-level BOUT++ settings such as the
 number of output timesteps ``nout`` and the output timestep size
 ``timestep`` (in normalised units where 95788 is one milisecond).
 Note that the solver timestep is adaptive and not user-settable.
 
 This is followed by ``[mesh]``, ``[solver]`` and ``[hermes]`` headers, where the ``[hermes]``
-section defines the list of components used. The component order
-matters, as the components are executed in order.
+section defines the list of components used. The component order doesn't
+matters, as they will be sorted to ensure state variables are set
```
mikekryjak, line 23:
```suggestion
matter, as they will be sorted to ensure state variables are set
```

include/permissions.hxx
mikekryjak, line 328:
```suggestion
/// FIXME: Currently these permissions are not expressed properly, due
```

src/component_scheduler.cxx
```diff
@@ -1,13 +1,432 @@
...
+/// permission by at least one component will be present.
+std::map<std::string, std::set<std::string>>
+getVariableHierarchy(const std::vector<std::unique_ptr<Component>>& components) {
+  // Build up a set of all variable names which are read only if they
+  // are set by another component
+  std::set<std::string> conditional_names;
+  for (const auto& component : components) {
+    const Permissions& permissions = component->getPermissions();
+    for (const auto& [varname, _] :
+         permissions.getVariablesWithPermission(PermissionTypes::ReadIfSet, true)) {
```
mikekryjak, line 169: I found this section hard to understand because of the bool. Can you split it into `getVariablesWithMinimumPermission` and `getVaraiblesWithPermission` instead?

src/component_scheduler.cxx
```diff
@@ -1,13 +1,432 @@
...
+    }
+  }
+
+  /// Assemble the set of all section names which are referred to
+  /// explicitly in the component permissions.
+  std::set<std::string> sections_present;
+  std::set_intersection(unconditional_names.begin(), unconditional_names.end(),
+                        unconditional_sections.begin(), unconditional_sections.end(),
+                        std::inserter(sections_present, sections_present.begin()));
+  /// Assemble the set of all variable names which are definitly
```
mikekryjak, line 194:
```suggestion
  /// Assemble the set of all variable names which are definitely
```

include/permissions.hxx
cmacmackin: You're right, it does not.

src/component.cxx
cmacmackin: You're right. There's actually a previous comment which needs to be changed too.

tests/unit/test_component.cxx
cmacmackin: I think there are some other functions which should be tested like this too.

bendudson (approved):
Thanks @cmacmackin !

src/component_scheduler.cxx
```diff
@@ -1,13 +1,432 @@
...
+  for (const auto& name : conditional_names) {
+    result.insert({name, {}});
+  }
+
+  // Non-sections map to themselves
+  for (const auto& name : non_sections) {
+    result[name] = {name};
+  }
+  // Sections map to those variables which they contain
+  for (const auto& section : sections_present) {
```
mikekryjak, line 215: Claude flagged an issue here and I've spent quite a long time trying to figure out whether it's a bug or not. I've spent a couple of hours trying to untangle it but I'm still not 100% sure. I think this function would benefit from a clearer docstring at the top going through the algorithm more clearly and in a bit more detail. It's good you have comments at each of the step but I'm finding it tricky to put them all together in my head even though I know what the function is for.

I can see that you are making a tree of all the permissions. The complaint is that you are making the tree by initially looking at paths of `unconditional_names` only. Then you make sure to account for `conditional_names` as well, but you fill the map with blanks which are only filled if the right permissions exist in `unconditional_names`. 

See a summary from Claude:

```
The hierarchy is built from `unconditional_names` only (L177-199). `conditional_names`
get an entry at L206-208, but an empty one — nothing in this function fills those. They
end up populated only if the same path independently arrives via the unconditional pile
at L212/L220.

So a `readIfSet` claim resolves correctly only when some component claims that exact
same path with `readOnly` or stronger. Whenever reader and writer have to be matched
through the tree, it fails — in both directions:

    Reader readIfSet("a:b")     Writer readWrite("a:b:c")
    Reader readIfSet("a:b:c")   Writer readWrite("a:b")

In each case the `readIfSet` name keeps its empty entry, `expandVariableName` returns
nothing, no dependency is created, and the reader may sort before the writer — silently.
Swap the reader to `readOnly` and both order correctly, which is what makes this a bug
rather than a limitation.

Fix: run the parent scan and the section/variable split over
`conditional_names ∪ unconditional_names`, then drop the L206-208 loop — every
conditional name then gets its entry from one of the other two loops, which
`expandVariableName`'s `.at()` requires.
```

src/component_scheduler.cxx
```diff
@@ -74,6 +493,8 @@ ComponentScheduler::ComponentScheduler(Options& scheduler_options,
   for (auto& component : components) {
     component->declareAllSpecies(species);
   }
+
+  ::sortComponents(components);
```
mikekryjak, line 497: We definitely need to have a flag to enable or disable the component sorting, as well as a console print at construction to print the input file component order and then the sorted list of components. I know you said this was in a separate PR - I'm assuming it's https://github.com/boutproject/hermes-3/pull/445? How come it's not in this PR? It opens a risk of issues between this and the next one being merged.

I'm sure you've explained this to me before, apologies for not remembering all of this. It's definitely good to get this merged soon.

mikekryjak (changes requested):
I gave another serious go at going through everything and left a few comments. My preference would be that we merge this with a flag to disable the sort if needed (enabled by default is fine) as well as a facility to print the as-written and as-sorted component list to console on initialisation.

We also definitely need to make a release before merging this in.

src/component_scheduler.cxx
cmacmackin: I'll have a think about what the correct behaviour should be here. To be honest, I don't really like it if people give write permission for whole sections, because you can't actually guarantee they'll have written every variable in them that another component might use. I think the only reason I allowed permissions to be set for whole sections was so that components could read all the collision rates for a species without having to work out in advance _exactly_ which ones were being computed.

cmacmackin: >I think this function would benefit from a clearer docstring at the top going through the algorithm more clearly and in a bit more detail.

Note that the overall algorithm is set out before the [sortComponents](https://github.com/boutproject/hermes-3/pull/428/changes#diff-3539d3bec156fbfbca693b2bf24246d4f38ed24a307d4e8327bea217d8e7d697R311) function, as well as in the documentation.

tests/unit/test_component_scheduler.cxx
```diff
@@ -67,3 +97,295 @@ TEST(SchedulerTest, SubComponents) {
...
+    TopologicalSort, ComponentOrderTest,
+    testing::Values(
+        // Empty set of components
+        Parameter({{"components", ""}}, {}),
+        // Single component with no permissions
+        Parameter(
+            {{"components", "a"},
+             {"a", {{"type", "orderchecker"}, {"permissions", toString(Permissions())}}}},
+            {"a"}),
+        // Single component with only read permissions
```
cmacmackin:
```suggestion
        // Single component with only write permissions
```

PR #432, opened 2025-12-02
parallel_ohms_law: Added a new component.

malamast:
1. Uses parallel Ohm's law to calculate the parallel current
2. Then uses this parallel current to calculate the parallel electron velocity.
3. It can potentially help with numerical stability in 2D asymmetric simulations (Tokamak plasma edge sims) when drifts are included. 
4. This component requires collisions to be calculated first, so it should be called after collisions component.
5. When parallel_ohms_law is used then there is no need to evolve_momentum for electrons

/// Caution! This component is not carfully evaluated for multispecies cases with multiple ions.

/// Use parallel Ohm's law to calculate the parallel current. 
/// This law basically comes from the electron parallel momentum equation 
/// by neglecting inertial (small by m_e / m_i) and viscous terms 
/// (see Phys. Plasmas 10, 4744–4757 (2003) by Simakov & Catto).
/// Use the parallel current to calculate the electron parallel velocity
///
///   Jpar = - 1/eta * ∇phi + 1/(eta e n_e) * ∇P_e + (0.71 k_B)/(eta e) * ∇T_e  
///
/// where phi is the electric potential and eta is the resistivity.
///
/// After normalization the equation takes the form: 
///
///   Jpar = - ∇phi / eta  +  ∇P_e / (eta n_e) + 0.71 ∇T_e / eta 
///
/// where the normalization parameter for the resistivity eta is:
///
///   eta_norm = Tnorm / (rho_s0 * Nnorm * Cs0 * qe)
///
/// Then uses this parallel current to calculate the parallel electron velocity.
///
/// Note: This needs to be called after collisions and other
///       components which impose forces on electrons_

### DIFF
--- a/src/parallel_ohms_law.cxx
+++ b/src/parallel_ohms_law.cxx
@@ -7,6 +7,34 @@
 
 using bout::globals::mesh;
 
+namespace {
+/// Limited free gradient of log of a quantity
+/// This ensures that the guard cell values remain positive
+/// while also ensuring that the quantity never increases
+///
+///  fm  fc | fp
+///         ^ boundary
+///
+/// exp( 2*log(fc) - log(fm) )
+///
+BoutReal limitFree(BoutReal fm, BoutReal fc) {
+  if (fm < fc) {
+    return fc; // Neumann rather than increasing into boundary
+  }
+  if (fm < 1e-10) {
+    return fc; // Low / no density condition
+  }
+  BoutReal fp = SQ(fc) / fm;
+#if CHECKLEVEL >= 2
+  if (!std::isfinite(fp)) {
+    throw BoutException("SheathBoundary limitFree: {}, {} -> {}", fm, fc, fp);
+  }
+#endif
+
+  return fp;
+}
+} // namespace
+
 ParallelOhmsLaw::ParallelOhmsLaw(std::string name, Options& alloptions, Solver*)
     : name(name) {
   AUTO_TRACE();
@@ -15,10 +43,10 @@ ParallelOhmsLaw::ParallelOhmsLaw(std::string name, Options& alloptions, Solver*)
     .doc("Save additional output diagnostics")
     .withDefault<bool>(false);
 
-  resistivity_floor = options["resistivity_floor"].doc("Minimum resistivity floor").withDefault(1e-4);
+  resistivity_floor = options["resistivity_floor"].doc("Minimum resistivity floor").withDefault(1e-5);
 
-  spitzer_resist = 
-        options["spitzer_resist"].doc("Use Spitzer resistivity?").withDefault<bool>(false);
+  spitzer_resistivity = 
+        options["spitzer_resistivity"].doc("Use Spitzer resistivity?").withDefault<bool>(false);
 
 
   const Options& units = alloptions["units"];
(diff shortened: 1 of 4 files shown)


mikekryjak:
Hi @malamast, nice work. Can you tell me a bit more about how mean field drifts with this approach compare to what SOLPS, UEDGE etc are doing? It looks more or less like what I understand is in SOLPS, but I'm not a drifts expert and never looked at the equations in detail.

src/parallel_ohms_law.cxx
```diff
@@ -0,0 +1,388 @@
...
+
+#include "../include/parallel_ohms_law.hxx"
+#include <bout/constants.hxx>
+#include "../include/hermes_utils.hxx"
+
+using bout::globals::mesh;
+
+
+namespace {
+BoutReal clip(BoutReal value, BoutReal min, BoutReal max) {
```
ZedThree: This is unused and can be deleted

src/parallel_ohms_law.cxx
```diff
@@ -0,0 +1,388 @@
...
+    Field3D LnLambda = 24.0 - log(sqrt(Zi * Ne_lim * Nnorm / 1.e6) / (Te * Tnorm)); // ln(Lambda)
+
+    BoutReal FZ = (1. + 1.198 * Zi + 0.222 * Zi * Zi) / (1. + 2.996 * Zi + 0.753 * Zi * Zi);  // correction coefficient in Spitzer model
+
+    resistivity_eta = Zeff * FZ * 1.03e-4 * Zi * LnLambda * pow(Te * Tnorm,-1.5); // eta in Ohm-m.
+
+  }
+
+  eta = softFloor( resistivity_eta / eta_norm , resistivity_floor );
+  // eta = sqrt( SQ(resistivity_eta / eta_norm) + resistivity_floor*resistivity_floor ); // This mitigates discontinuities.
```
ZedThree: We can delete commented out code:
```suggestion
```

ZedThree (changes requested):
Some little bits to tidy up, but otherwise looks good. I'll leave the physics review up to @mikekryjak / @bendudson !

malamast:
Lets keep this open. Hey @tbody-cfs.  Is there a way to extend the 90 days period ?

tbody-cfs:
> Lets keep this open. Hey @tbody-cfs. Is there a way to extend the 90 days period ?

@malamast this PR will be marked as non-stale now that you've commented. Generally speaking, in the dev meeting it was decided that it's acceptable to give an update comment.

The stale tag will be auto-removed when the action next runs (either a cron job or via a workflow dispatch). I triggered a run in https://github.com/boutproject/hermes-3/actions/runs/29113383513/job/86430800355, and it looks like this successfully removed the stale tag due to your comment.

bendudson:
Hi @malamast ! Sorry, I will review this asap. I think commenting on the PR, or removing the `stale` label, should keep the PR going for another 90 days. Hopefully it will be merged before then.

malamast:
> Hi @malamast ! Sorry, I will review this asap. I think commenting on the PR, or removing the `stale` label, should keep the PR going for another 90 days. Hopefully it will be merged before then.

No, that's fine.  It's still WIP. I just wasn't sure how to remove the stale label.

PR #436, opened 2025-12-09
Refactor permission substitution to do all substitutions at once

cmacmackin:
This was a suggestion made by @ZedThree when review #421. I thought I saw a way to do this but it proved to be quite involved. By the time I realised that I was most of the way done so I decided to keep the work on a separate branch just in case we wanted to use it for something. However, I think this approach is actually less flexible and ultimately more complex. For example, it is now harder to predict the order in which substitutions occur. This means that if there are two substitutions that will ultimately result in the same variable name we don't know which one will be kept.

### DIFF
--- a/include/component.hxx
+++ b/include/component.hxx
@@ -3,15 +3,6 @@
 #ifndef HERMES_COMPONENT_H
 #define HERMES_COMPONENT_H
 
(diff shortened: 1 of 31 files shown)


cmacmackin:
It was agreed that this didn't make sense as a change.

Issue #467, opened 2026-01-14
Many warnings of permissions which are too loose


mikekryjak:
Many of the permissions are currently more open than required, creating a lot of warnings, e.g.:

```
Did not read from state variable species:d:density in region(s) , , Interior
Did not write to state variable species:d:momentum_source in region(s) , , Interior, Boundaries
Did not write to state variable species:e:momentum_source in region(s) , , Interior, Boundaries
Did not write to state variable species:d+:pressure in region(s) , , Boundaries
Did not write to state variable species:e:pressure in region(s) , , Boundaries
```

These should be tightened up so they are not so numerous. However, this is hard because the components don't know their own name, so the warning can't identify where it came from. Because of this, we need to resolve https://github.com/boutproject/hermes-3/issues/429 to have a chance at fixing this. In the meantime the warnings should be suppressed to allow high `DCHECK` levels to be useful.

Issue #429, opened 2025-11-28
Components should know their name


ZedThree:
Not all components store their name -- the base class should handle this

PR #468, opened 2026-01-14
Suppress permission warnings

mikekryjak:
There are warnings when permissions are too loose (https://github.com/boutproject/hermes-3/issues/467). They are very numerous and hard to fix because they don't identify the source component. We need to add the capability for components to know their name to have a realistic chance of resolving this (https://github.com/boutproject/hermes-3/issues/429). 

The warnings are so large that the logs quickly grow to multi-MB in size, making it impractical to run the code or see other checks. Because of this, I think we should disable the warnings until we have a chance of fixing them. 

This PR does this by requiring a `DCHECK=4` (which doesn't exist).

Permissions being too open will not be a problem until the topological sort algorithm is finished (https://github.com/boutproject/hermes-3/pull/428), at which point they will inform the component order. We can resolve this when we work on the algorithm.

### DIFF
--- a/src/component.cxx
+++ b/src/component.cxx
@@ -20,7 +20,7 @@ std::unique_ptr<Component> Component::create(const std::string &type,
 void Component::transform(Options& state) {
   GuardedOptions guarded(&state, &state_variable_access);
   transform_impl(guarded);
-#if CHECKLEVEL >= 4
+#if CHECKLEVEL >= 999
   for (auto& [varname, region] : guarded.unreadItems()) {
     output_warn.write("Did not read from state variable {} in region(s) {}\n", varname,
                       Permissions::regionNames(region));
--- a/src/component.cxx
+++ b/src/component.cxx
@@ -20,7 +20,7 @@ std::unique_ptr<Component> Component::create(const std::string &type,
 void Component::transform(Options& state) {
   GuardedOptions guarded(&state, &state_variable_access);
   transform_impl(guarded);
-#if CHECKLEVEL >= 1
+#if CHECKLEVEL >= 3
   for (auto& [varname, region] : guarded.unreadItems()) {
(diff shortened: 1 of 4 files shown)


src/component.cxx
```diff
@@ -20,7 +20,7 @@ std::unique_ptr<Component> Component::create(const std::string &type,
 void Component::transform(Options& state) {
   GuardedOptions guarded(&state, &state_variable_access);
   transform_impl(guarded);
-#if CHECKLEVEL >= 1
+#if CHECKLEVEL >= 3
```
cmacmackin:
```suggestion
#if CHECKLEVEL >= 4
```
Because this is `>=`, your initial change would have made these CHECKLEVEL 3.

cmacmackin:
This will be enough to suppress the warnings messages, but if we're turning those off there are some other bits it would make sense to turn off too. See https://github.com/oparry-ukaea/hermes-3/commit/5e66efb9b9774ed95a514cf2df7d37839c52a555.

cmacmackin:
>I always run with CHECK=4. If you think this should currently be disabled, use something like 999 to make the intent clearer, or even better remove it. You can then later revert it, if this becomes useful.

I don't think `-DCHECK=4` is meant to do anything. The [BOUT++ docs only mention check levels up to 3](https://bout-dev.readthedocs.io/en/stable/user_docs/advanced_install.html#optimisation-and-run-time-checking) and there are only assertion macros up to that level too.

dschwoerer:
They are still printed with debug level 4.

You could also print them to `output_debug` instead of `output_warn` - if you want the output of by default.

Anyway, this should not be merged as it decreases test coverage :innocent:

bendudson (approved):
I think this is ok for now, to make `master` usable. Setting `CHECK` to 999 allows testing of this feature before merging.

PR #471, opened 2026-01-16
parallel inertia curvature drift

malamast:
Work in progress. 

I included this curvature drift.  It's effect is probably small but it is usually included in 2D transport simulations (UEDGE and SOLPS). 
This term comes from the parallel inertia term in the momentum equation.
It is due to parallel inertia in a curved magnetic field (centrifugal-like force). 
See **~Vpar^2** term in the paper by A.V. Chankin (Journal of Nuclear Materials 241-243 (1997) 199-213) 

**Any ideas for a better name for this component**? It could also be included in the diamagnetic_drift since the calculations are very similar and both drifts require Curb_B. 

I calculate the divergence of the current due to this drift and add it to **DivJextra** so there is no need to modify the vorticity component. 

      components = (d+, e, d, 
                    braginskii_collisions, vorticity,
                    sheath_boundary_simple,  
                    braginskii_friction, braginskii_heat_exchange,
                    sound_speed, reactions, 
                    braginskii_ion_viscosity,
                    diamagnetic_drift, parallel_inertia_curvature_drift,
                    braginskii_thermal_force, braginskii_conduction, recycling)


malamast:
The evolve_pressure file was also accidentally changed. I'll have to fix that.

PR #473, opened 2026-01-19
Add 2D integrated test to CI

mikekryjak:
This PR adds an integrated test based on Test 2 from the benchmark suite (https://github.com/mikekryjak/hermes-perftest). 
The simulation is from my upcoming paper on the SOLPS and SOLEDGE2D comparison which is representative of a production simulation without viscosity, impurities, currents or drifts. It's a double null equilibrium in an attached, H-mode example future scenario of ST40.

The test downloads the grid and restart files from Zenodo and runs the simulation for **five timesteps of 0.0001 each** using `--oversubscribe`. It compares the values of `Pe` and `NVd+` at all four targets, the SOL boundary, the core/PFR boundaries and the first SOL ring. If the test fails, it will print which region and variable had the greatest error. My thinking is that `Pe` will capture both density and energy and `NVd+` will capture momentum.

- [x] Create the test and run locally
- [x] Upload grid and restart files to Zenodo
- [x] Add homemade download code
- [x] Make test actually run the simulation
- [x] Add to CI
- [x] Add docs

Questions:
- Should/Can the test zip file be downloaded using CMake? @oparry-ukaea I remember you saying that it can only download one file - I may be mistaken.


bendudson (approved):
Thanks @mikekryjak !

PR #482, opened 2026-01-23
Fix sheath boundary simple 2

bendudson:
When `phi` is not set, `phisheath` is calculated from the ion flux. If `nesheath` or `ion_sum` go to zero then phisheath becomes NaN. That caused vesheath to be NaN, then the electron heat flux. Simulations would fail due to an invalid value being assigned to `e:energy_source`.

This applies floors to `nesheath` and `ion_sum` in the calculation of phi. [Note: Possibly should be `softFloor`?]

Replaces #478

Not covered by unit tests but fixes an issue with production simulations. The sheath boundary components are being consolidated and tested by @ZedThree .


bendudson:
Bypassing coverage tests because this fixes a problem with production simulations, and the sheath boundaries are in the process of being replaced.

PR #485, opened 2026-01-26
CI: Add clang-format for changes

### LINKED ISSUE #360: We should have clang-format in CI
As per the title. It's currently only applied by Ben - I didn't actually know we were supposed to do that until recently!

dschwoerer:
The file is copied from the BOUT-dev repo.

Co-Author: Peter Hill <peter.hill@york.ac.uk>

Resolves: #360

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,33 @@
+name: clang-format-command
+
+on:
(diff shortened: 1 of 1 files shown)


bendudson (approved):
Thanks @ZedThree ! I agree with using the same process as BOUT++.

Issue #325, opened 2025-04-07
Neutral pump implemented at targets


mikekryjak:
The neutral pump is currently only implemented for the SOL and PFR. 
While it's not feasible to put a turbopump or cryopump at the target, the target surface can act as a pump, in particular in carbon devices such as MAST-U where the wall absorbs the plasma. 

I think this would improve our match to SOLPS @PoloidalLloyd. I have to say I forgot that we didn't have this.

bendudson:
Usually this would be handled by setting a recycling coefficient less than 1. I guess a pump would work in a different way, removing neutrals from regions where there may not be a plasma, but I suspect the effect will be much the same.

mikekryjak:
Hi @bendudson indeed, this is how the pump is implemented on the radial boundaries at the moment, with a "recycling factor" (neutral albedo). I just need to extend this to the targets as well.

PR #491, opened 2026-01-30
Add target pump

mikekryjak:
This PR solves https://github.com/boutproject/hermes-3/issues/325. It adds the capability for target neutral pumps in the same way as the SOL and PFR neutral pumps. It also changes the Sd/Ed_pump diagnostics to refer to the source just due to the pump, recycling minus pump.

It also modifies the 2D-production test to have a pump at the target, and adds a Python reproduction test of the target pump to the 2D-recycling test.

Tasks:

- [x] Implement for upper boundary
- [x] Implement for lower boundary
- [x] Change diagnostics so that Sd_pump and Ed_pump are due to the pump only, not pump + recycling
- [x] Check impact on 2D-production result
- [x] Update 2D-recycling and 2D-production tests
- [x] Update documentation
- [x] Remove debug variables
- [x] Fix missing assert on outer lower recycling test
- [x] Remove unnecessary energy_flow_ylow read

The 2D-recycling test saves a handy plot to check whether the reconstruction of the 8 fields is correct. It has the ability to plot debug variables so you can easily rectify any future discrepancies without needing a separate script.

image


src/recycling.cxx
```diff
@@ -284,16 +277,59 @@ void Recycling::transform_impl(GuardedOptions& state) {
...
+                  * channel.target_fast_recycle_fraction // Fast recycling part
+              + recycle_particle_flow * (1 - channel.target_fast_recycle_fraction)
+                    * channel.target_energy; // Thermal recycling part
+
+          // Calculate neutral pump neutral sinks due to neutral impingement
+          BoutReal pump_neutral_particle_sink = 0.0;
+          BoutReal pump_neutral_energy_sink = 0.0;
+
+          // These are additional to particle sinks due to recycling
+          if ((is_pump[i] == 1.0) and (neutral_pump)) {
```
bendudson:
```suggestion
          if ((is_pump[i] > 0.5) and (neutral_pump)) {
```

src/recycling.cxx
```diff
@@ -267,12 +258,14 @@ void Recycling::transform_impl(GuardedOptions& state) {
                                  * (dx[i] + dx[ig]) * 0.5 * (dz[i] + dz[ig]);
           BoutReal volume = J[i] * dx[i] * dy[i] * dz[i];
 
-          // Flow of recycled neutrals into domain [s-1]
-          BoutReal flow = channel.target_multiplier * flux * daparsheath;
+          // If cell is a pump, overwrite multiplier with pump multiplier
+          BoutReal multiplier = channel.target_multiplier;
+          if ((is_pump[i] == 1.0) and (neutral_pump)) {
```
bendudson:
```suggestion
          if ((is_pump[i] > 0.5) and (neutral_pump)) {
```

bendudson (approved):
Thanks @mikekryjak ! Some minor suggestions to avoid float comparisons.

PR #493, opened 2026-02-04
Upgrade to Spack v1.1.0 and latest VANTAGE dev branch

mikekryjak:
- [x] Merge latest Hermes-3 master
- [x] Update VANTAGE and other submodules
- [x] Fix Spack v1 config file syntax changes
- [x] Concretization successful
- [ ] Install successful
- [ ] Update particle pusher app


.github/workflows/tests.yml
```diff
@@ -3,10 +3,10 @@ name: Tests
 on:
   push:
     branches:
-      - master
+      - '**'
   pull_request:
     branches:
-      - '*'  # This will trigger for PRs against any branch
+      - '**'  # This will trigger for PRs against any branch
```
dschwoerer: Why `'**'`?

I think this could be just `on: [push, pull_request]`

dschwoerer (commented):
I am not sure we should start including python scripts for post-processing.
Should they maybe go to boutdata, or maybe to its own repository?
That would also allow to put the tools on pypi.

Issue #490, opened 2026-01-30
Bug in fast recycling & component order problem


mikekryjak:
In fast recycling, the returning neutral keeps some of the energy of the incident ion.
This means the `recycling` component must know the sheath ion heat flux. When I implemented this, I used `energy_flow_ylow` to get this:
https://github.com/boutproject/hermes-3/blob/f6ae2d1e7838218a13aeb1dfea61cc6704f26365/src/recycling.cxx#L290-L291

Heat flux through the sheath has three components:
- Pressure advection in `finally()` of `evolve_pressure`
- Conduction in `transform()` of `braginskii_conduction`
- Extra sheath heat flux in `transform()` of `sheath_boundary_simple`  

All three must find their way into `energy_flow_ylow` in the state to be picked up by `recycling`.
There is a bug where `evolve_pressure` is not writing its contribution into the state. This means that right now, the fast recycling energy is being calculated from only the other contributions, which can even be the wrong sign depending on the sheath heat transmission coefficient choice!

I started trying to fix this before I realised that it may be impossible - the state cannot be modified in `finally()`, which means `evolve_pressure` cannot save its advective flow into the state!

This calls for a rethink of using `energy_flow_ylow` for something other than diagnostics. And a think about how we can get that advection term into `recycling`.

PR #494, opened 2026-02-05
Fix fast recycling heat flux

mikekryjak:
Previously, the thermal/fast recycling calculation relied on `energy_flow_ylow` to know the incident ion energy flux on the target. Unfortunately, this only captured the "additional" heat flux from `sheath_boundary_simple` to get the heat flux to the user-set gamma, while the main component is the advected heat flux from `evolve_pressure`. This issue is described in https://github.com/boutproject/hermes-3/issues/490 and is resolved by this PR.

Unfortunately it is not possible to obtain the advected heat flux from the state, because `evolve_pressure` calculates it in `finally()`. This means the sheath heat flux must be reconstructed in `recycling` using the known Ni, Ti, Vi at the sheath along with a known `gamma_i`. Since currently only `sheath_boundary_simple` has a `gamma_i` setting, this means fast recycling is incompatible with the other sheath boundaries, but this was already the case before this PR. This can be resolved as part of https://github.com/boutproject/hermes-3/pull/382, where all the sheath boundaries will be consolidated allowing one to obtain a `gamma_i` in a reasonable way.

Tasks
- [x] Implement the fix
- [x] Perform a minor refactor to use simpler indexing, e.g. Ne[i] + Ne[g] instead of N(r.ind, mesh->ystart, jz) + N(r.ind, mesh->ystart - 1, jz). This makes the code nearly identical between the upper and lower boundaries and reduces the chance of typos. 
- [x] Test the impact of the changes on the 2D production test
- [x] Add the first of several planned tests around boundaries which reconstruct the C++ in Python. 

The impact on the 2D-production test is small and mostly to do with neutral conditions at the target. The neutral temperature increases because the fast recycling of energy is now correct:
image


bendudson (approved):
Thanks @mikekryjak !

Issue #503, opened 2026-02-11
Permissions bug in tokamak-2D-driftplane/2D-drift-plane-turbulence-te-ti


jvwilliams23:
Maybe duplicate of #498 or #484 

I get a permissions bug in the example `tokamak-2D-driftplane/2D-drift-plane-turbulence-te-ti`. Here is the log file:

```
Error encountered: Stack trace (most recent call first):
#0  (filtered)
#1  (filtered)
#2  (filtered)
#3  in Permissions::checkNoRemainingSubstitutions() const
    at /work4/cse/scarf1324/hermes-3/src/permissions.cxx:116:55
     114:   if (unsubstituted.size() > 0) {
     115:     throw BoutException("The following variable names have unsubstituted labels: {}",
   > 116:                         fmt::join(unsubstituted, ", "));
                                                                ^
     117:   }
     118: }
#4  in ComponentScheduler::ComponentScheduler(Options&, Options&, Solver*)
    at /work4/cse/scarf1324/hermes-3/src/component_scheduler.cxx:78:33
      76:
      77:   for (auto& component : components) {
    > 78:     component->declareAllSpecies(species);
                                          ^
      79:   }
      80: }
#5  (inlined)          in std::__detail::_MakeUniq<ComponentScheduler>::__single_object std::make_unique<ComponentScheduler, Options&, Options&, Solver*&>(Options&, Options&, Solver*&)
                       at /apps20/sw/easybuilt/rocky/9/amd/zen2/software/GCCcore/14.2.0/include/c++/14.2.0/bits/unique_ptr.h:1076:30
    1074:     inline __detail::__unique_ptr_t<_Tp>
    1075:     make_unique(_Args&&... __args)
  > 1076:     { return unique_ptr<_Tp>(new _Tp(std::forward<_Args>(__args)...)); }
                                       ^
    1077:
    1078:   /** Create an array owned by a `unique_ptr`.
#6  in ComponentScheduler::create(Options&, Options&, Solver*)
    at /work4/cse/scarf1324/hermes-3/src/component_scheduler.cxx:86:72
      84:                                                                Solver *solver) {
      85:   return std::make_unique<ComponentScheduler>(scheduler_options,
    > 86:                                               component_options, solver);
                                                                                 ^
      87: }
#7  in Hermes::init(bool)
    at /work4/cse/scarf1324/hermes-3/hermes-3.cxx:288:41
     286:   // Options::root() is passed as the root of the component options, so that
[...]

PR #509, opened 2026-02-15
sheath_closure: fix permissions typo

### LINKED ISSUE #498: Permissions bug in classical_diffusion
The examples/tokamak-2D/turbulence example doesn't run because classical_diffusion is missing AA permissions.

mikekryjak:
Solves https://github.com/boutproject/hermes-3/issues/503.

I also wanted to fix https://github.com/boutproject/hermes-3/issues/498 but I can't seem to reproduce it for some reason.


ZedThree (approved):

mikekryjak:
> Maybe we need a really dumb test like "try to create each component call `transform` on it" just to check its permissions are set correctly?
> 
> Although I guess each test might need special set up to ensure it has the correct state set?

That would probably make sense. At the moment this is done for several components as a unit test, but not for all.

PR #512, opened 2026-02-20
Add Testing dir to gitignore

mikekryjak:
Adds `**/Testing/**` to .gitignore. This is a directory created when tests are run locally.


dschwoerer:
Should the testing not be in the build folder, which is anyway ignored? How do you get those files?

If that is actually needed, can you end the file with a new line?

bendudson (approved):

PR #516, opened 2026-02-24
Fix some sphinx warnings

oparry-ukaea:
Get rid of some sphinx warnings.
Think this covers everything other than the "Duplicate C++ declaration, also defined at..." warnings from breathe. 
Feedback welcome!


mikekryjak:
Wonder if we can make the warnings errors? Is it possible to resolve all of them?

mikekryjak:
Ok - let's revisit once that's fixed then.

PR #451, opened 2025-12-30
Neutrals flux limiters

malamast:
Just a few suggestions for neutral_mixed and other components:

1) added a functionality to initialize n and p from grid variables. 
2) Changed the way we calculate the flux limiters and transport coefficients of neutrals. I separate the gradients for the flux limiters in parallel and perpendicular directions. They do something similar in uedge as well. 
3) Added the contribution of ion velocities to the neutral perpendicular velocity. This term comes for momentum sources in the derivation of the neutral perpendicular velocity. How can we generalize it here for multiple species?
4) minor changes to zero_current and braginskii_collisions components.

**MK edits:**

I am taking the PR over from Malamas and finishing it. Tasks:
- [x] Changes from Malamas: separate limiters for parallel/perpendicular conductivity and viscosity, perpendicular ion velocity coupling, and so on
- [x] Change limiter formulation to AFN style
- [x] Add user settable limiters for each flux
- [x] Change free streaming velocities to AFN definition
- [x] Remove double-counting of `neutral_lmax`
- [x] Make perpendicular ion velocity coupling a user option
- [x] Test against historical SOLPS comparison results
- [ ] Update 2D-Production test
- [ ] Add flux limiter reconstruction to 2D-recycling and rename to 2D-reconstruct
- [ ] Documentation

PR #530, opened 2026-03-12
Added floors to several components. 

malamast:
-That is to protect some variables from becoming Nan when the SNES solver increases the timestep aggressively. 
-ion_viscosity was causing problems.
-In the hydrogen_charge_exchange I changed the loop to include boundaries because later we take log(Teff) and if the boundaries of Teff become negative we could potentially have a problem? not sure about that. 
-In the sheath_boundary_simple we had two return statements in that function so I removed the first one. In that way, the check can take effect. 
-I removed the density2 variable in braginskii_friction because it was not used.
-In the electron_viscosity i included even more floors (compared to the ion_viscosity) because this component was causing also many problems with snes. 
-Added a isfinite check of the time derivatives  in vorticity and relax_potential components as we do with the rest of the components.


mikekryjak:
I am breaking up https://github.com/boutproject/hermes-3/pull/451 into separate PRs and would like to put the floor changes into this branch if that's OK.

src/braginskii_thermal_force.cxx
```diff
@@ -38,7 +38,7 @@ void BraginskiiThermalForce::transform_impl(GuardedOptions& state) {
 
       const BoutReal Z = get<BoutReal>(species["charge"]);
       // Don't need density boundary
-      const Field3D nz = GET_NOBOUNDARY(Field3D, species["density"]);
+      const Field3D nz = floor(GET_NOBOUNDARY(Field3D, species["density"]), 0.0);
```
bendudson:
```suggestion
      const Field3D nz = GET_NOBOUNDARY(Field3D, species["density"]);
```

src/braginskii_thermal_force.cxx
```diff
@@ -123,7 +123,8 @@ void BraginskiiThermalForce::transform_impl(GuardedOptions& state) {
 
         const BoutReal mz = get<BoutReal>((*heavy)["AA"]);
         const BoutReal Z = get<BoutReal>((*heavy)["charge"]);
-        const Field3D nz = GET_NOBOUNDARY(Field3D, (*heavy)["density"]);
+        const Field3D nz = floor(GET_NOBOUNDARY(Field3D, (*heavy)["density"]), 0.0);
```
bendudson:
```suggestion
        const Field3D nz = GET_NOBOUNDARY(Field3D, (*heavy)["density"]);
```

src/hydrogen_charge_exchange.cxx
```diff
@@ -44,8 +44,8 @@ void HydrogenChargeExchange::calculate_rates(GuardedOptions&& atom1, GuardedOpti
   // Optionally multiply by arbitrary multiplier
   const Field3D sigmav = exp(ln_sigmav) * (1e-6 * Nnorm / FreqNorm) * rate_multiplier;
 
-  const Field3D Natom = get<Field3D>(atom1["density"]);
-  const Field3D Nion = get<Field3D>(ion1["density"]);
+  const Field3D Natom = softFloor(get<Field3D>(atom1["density"]), 0.0);
```
bendudson:
```suggestion
  const Field3D Natom = get<Field3D>(atom1["density"]);
```

src/hydrogen_charge_exchange.cxx
```diff
@@ -44,8 +44,8 @@ void HydrogenChargeExchange::calculate_rates(GuardedOptions&& atom1, GuardedOpti
...
+  const Field3D Nion = softFloor(get<Field3D>(ion1["density"]), 0.0);
```
bendudson:
```suggestion
  const Field3D Nion = get<Field3D>(ion1["density"]);
```

bendudson (approved):

mikekryjak:
Can we incorporate the flooring changes from #451 into this PR?
I found:
- `braginskii_collisions` density floor no longer hardcoded: https://github.com/boutproject/hermes-3/blob/94974762e993c94ca47be07e72a6243167a8fdcc/src/braginskii_collisions.cxx#L149
- New floor in `zero_current`: https://github.com/boutproject/hermes-3/blob/94974762e993c94ca47be07e72a6243167a8fdcc/src/zero_current.cxx#L73

There are also some floors in `neutral_mixed` but I'm putting those into #556 where I'm implementing just the flux limiter bits of #451.

malamast:
Hi @mikekryjak and @bendudson,  what should we do with this? It seems its outdated now. I can make a new one of you want.

PR #533, opened 2026-03-24
Shorten timestep on 2D-production/recycling tests

mikekryjak:
Reduced the required timestep/nout for 2D-recycling and 2D-production to a very small amount.
On my system they take ~35s each with the vast majority of the time taken by xBOUT loading.

I also added some more verbosity and timing messages. Verbosity is now True by default. It's only a handful of lines and the extra timing information could come in handy.

We could get further speedups with the new, faster xBOUT loading (https://github.com/boutproject/xBOUT/pull/336) but this is still not yet merged due to yet another issue with xBOUT tests - this time it looks like it's an Xarray file lock race condition which shows up only in newer Python versions. 

Technically the tests could be rewritten to use `boutdata.collect`, but this would not be trivial since they use xHermes region selection logic.


tests/integrated/2D-production/runtest
```diff
@@ -8,9 +8,12 @@ import os
 from pathlib import Path
 import urllib.request
 import xhermes
+from time import time
+
+t0 = time()
 
 gen_data = False
-verbose = False
+verbose = True
```
oparry-ukaea: Should this default to True?
Doesn't seem like much reason to have this option if it's almost always on anyway?

oparry-ukaea (approved):

PR #534, opened 2026-03-25
Decouple reaction data

oparry-ukaea:
Decouples reaction data from the reaction classes themselves.

1. Reaction data type and src ID (e.g. Amjuel, "H.4_2.1.5") for each reaction can be set in the config file, but the defaults preserve the existing behaviour if nothing is specified.
2. The data objects are constructed using a factory.
3. The new reaction class hierarchy is:
  - Reaction
    - CXReaction 
    - RecIznReaction
      - RecReaction
      - IznReaction 
4. And the data class hierarchy is
  - ReactionData
    - ReactionDataWithCoeffs
      - AmjuelData
      - [ADASData]
      - ... [data providers that generate rates via functional fits]
    - ... [any other data provider] (e.g. a CRM code, run either in real time at startup, or via files in some new format.)

---

ToDo:
- [x] Rebase and split up commits
- [x] Update developer docs
- [x] Add some more tests

---

Not planned for this PR:
- Make it possible to configure which species are used for rate parameters, i.e. R(n_s, T_s), R(T_s) rather than R(n_e, T_e) and T(T_eff)
- Add ADASData and switch impurity reactions to new framework


docs/sphinx/reactions.rst
```diff
@@ -203,12 +203,16 @@ NOTE:
 Implemented reactions 
 ----------------------
 
-Hydrogen
-~~~~~~~~
+Hydrogen and Helium
```
mikekryjak, line 211: Am I right in thinking these are grouped together because they are AMJUEL rates as opposed to the impurities which are ADAS? If so it may be useful to add a note to this effect

mikekryjak (changes requested):
Looks great to me, it's satisfying to finally not have a file called `amjuel_hydrogen`. I only have a couple of small suggestions on the docs. In addition to the one I put down as a review comment, I think it would also be good to put a link to the Developer Manual section on reactions in the top level one. Ideally with an eye catching box with an exclamation mark or something like that.

docs/sphinx/reactions.rst
oparry-ukaea: It's mainly because they're using the new framework and the impurities aren't.
I think the species-level subsections here make less with a generic approach to reactions - I kept it in because it was a convenient place to list the default datasets!

oparry-ukaea:
> I think it would also be good to put a link to the Developer Manual section on reactions in the top level one. Ideally with an eye catching box with an exclamation mark or something like that.

Done.

bendudson (approved):

PR #537, opened 2026-03-28
Fci: Add 3D MMS tests and fix neutral averaging

totork:
This PR does add more MMS test, now in a more realistic 3D geometry. The magnetic field is axial and circular with a finite q-profile. The q profile can be varied radially, but is kept constant for these tests (q=3). There are 2 tests:

1. A simpler 3D tests created with a constant parallel diffusion. 
2. A 3D test that uses the electron heat conduction for the parallel diffusion. 

As the field is axial, the field strength on a field line stays constant, which means that it does not test flux expansion.

**These tests also install Zoidberg/better-metric in the CI. Building this PR requires a Zoidberg version, atleast for all the tests. I am not sure if the hermes-3 executable is built properly when zoidberg is not found.**


.github/workflows/tests.yml
```diff
@@ -57,6 +57,7 @@ jobs:
       - name: Install pip packages
         run: |
           ./.pip_install_for_tests.sh 'netcdf4~=1.5' 'boutdata==0.3.0' boututils
+          python -m pip install git+https://github.com/boutproject/zoidberg@better-metric
```
dschwoerer, line 60:
```suggestion
          ./.pip_install_for_tests.sh 'netcdf4~=1.5' 'boutdata==0.3.0' boututils 'git+https://github.com/boutproject/zoidberg@better-metric'
```

dschwoerer (commented):
It would be nice to remove the python2 fallback paths as well.

Otherwise, this seems pretty straight forward 👍

.github/workflows/tests.yml
```diff
@@ -56,7 +56,7 @@ jobs:
 
       - name: Install pip packages
         run: |
-          ./.pip_install_for_tests.sh 'netcdf4~=1.5' 'boutdata==0.3.0' boututils 'git+https://github.com/boutproject/zoidberg@better-metric'
+          ./.pip_install_for_tests.sh 'netcdf4~=1.5' 'boutdata==0.3.0' boututils 'git+https://github.com/boutproject/zoidberg'
```
dschwoerer:
```suggestion
          ./.pip_install_for_tests.sh 'netcdf4~=1.5' 'boutdata==0.3.0' boututils zoidberg
```
or 
```suggestion
          ./.pip_install_for_tests.sh 'netcdf4~=1.5' 'boutdata!=0.4.0' boututils zoidberg
```
so we get the next boutdata automatically.

totork: Last time I used boutdata=0.4.0, I would get weird errors and the CI would fail all over the place. I am not sure if this is the case anymore.

bshanahan: @dschwoerer can we mark this as resolved?

bshanahan:
test failure looks like an issue from the main `fci` branch where T doesn't have parallel slices properly in evolve_pressure.

dschwoerer (approved):
This looks good now 👍

PR #539, opened 2026-04-02
diamagnetic_drift: Boundary flux changes

bendudson:
### Purpose
Ensure that the divergence of the diamagnetic drift is zero, and that fluxes through the core boundary doesn't result in unphysical boundary layers.

### Change Summary
- Replaces blending of divergence and gradient form with a switch
- Adds a core averaging option that averages diamagnetic terms over the x surface next to the core boundary
- Refactors the `transform_impl` function into public functions
- Adds unit tests of public functions

### Validation
Unit tests of steps in the calculation, following refactor. Includes tests of the conservation properties.

### AI Assistance
Codex/GPT-5.4 wrote most of the code. I planned, reviewed and adjusted the changes.

### Documentation
The diamagnetic drift equation documentation has been updated.


src/diamagnetic_drift.cxx
```diff
@@ -1,36 +1,71 @@
...
-  diamag_form = options["diamag_form"]
-    .doc("Form of diamagnetic drift: 0 = gradient; 1 = divergence")
-    .withDefault(Field2D(1.0));
+  bndry_flux = options["bndry_flux"]
+                   .doc("Allow fluxes through boundary?")
+                   .withDefault<bool>(false);
+
+  divergence_form = options["divergence_form"]
+                        .doc("Use divergence form of diamagnetic drift?")
+                        .withDefault<bool>(true);
```
cmacmackin, line 37: Call this option "form" and use [BoutEnum](https://bout-dev.readthedocs.io/en/latest/user_docs/bout_options.html#types-for-multi-valued-options) to provide options with meaningful names (e.g., `divergence` and `gradient`).

src/diamagnetic_drift.cxx
```diff
@@ -69,50 +104,115 @@ DiamagneticDrift::DiamagneticDrift(std::string name, Options& alloptions,
...
+
+    subtract(species["momentum_source"], sink);
+  }
+}
+
+void DiamagneticDrift::coreAverage(Field3D& f) {
+  if (!average_core) {
+    throw BoutException("DiamagneticDrift::coreAverage requires average_core=true");
+  }
+
```
cmacmackin, line 188: We can avoid this check if we make `coreAverage` is a standalone function that take `core_ring_volume` and `cell_volume` as arguments.

src/diamagnetic_drift.cxx
```diff
@@ -69,50 +104,115 @@ DiamagneticDrift::DiamagneticDrift(std::string name, Options& alloptions,
...
 
-    if (!(species.isSet("charge") and species.isSet("temperature")))
-      continue; // Skip, go to next species
+Field3D DiamagneticDrift::calculateGradientForm(const Field3D& quantity,
+                                                const Field3D& temperature,
+                                                BoutReal charge, BoutReal factor) {
+  if (fabs(charge) < 1e-5) {
+    throw BoutException(
+        "DiamagneticDrift::calculateGradientForm requires a non-zero charge");
+  }
```
cmacmackin, line 125: Don't need this, as we already check the charge before calling.

src/diamagnetic_drift.cxx
```diff
@@ -69,50 +104,115 @@ DiamagneticDrift::DiamagneticDrift(std::string name, Options& alloptions,
...
-void DiamagneticDrift::transform_impl(GuardedOptions& state) {
-  // Iterate through all subsections
-  GuardedOptions allspecies = state["species"];
+Field3D DiamagneticDrift::calculateDivergenceForm(const Field3D& quantity,
+                                                  const Field3D& temperature,
+                                                  BoutReal charge, BoutReal factor) {
+  if (fabs(charge) < 1e-5) {
+    throw BoutException(
+        "DiamagneticDrift::calculateDivergenceForm requires a non-zero charge");
+  }
```
cmacmackin, line 113: Don't need this, as we already check the charge before calling.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+  const Field2D cell_volume = coords->dx * coords->dy * coords->dz * coords->J;
+
+  BoutReal result = 0.0;
+  for (int jx = mesh->xstart; jx <= mesh->xend; ++jx) {
+    for (int jy = mesh->ystart; jy <= mesh->yend; ++jy) {
+      for (int jz = mesh->zstart; jz <= mesh->zend; ++jz) {
+        result += cell_volume(jx, jy) * field(jx, jy, jz);
+      }
+    }
+  }
```
cmacmackin, line 52:
```suggestion
  BOUT_FOR(i, mesh->getRegion3D("RGN_NOBNDRY")) {
    result += cell_volume[i] * field[i];
  }
```

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+}
+
+class ScopedGlobalMesh {
+public:
+  explicit ScopedGlobalMesh(Mesh* new_mesh) : previous_mesh(mesh) { mesh = new_mesh; }
+  ~ScopedGlobalMesh() { mesh = previous_mesh; }
+
+private:
+  Mesh* previous_mesh;
+};
```
cmacmackin, line 63: Instead of adding a new class, which must be created within the body of a test, why not add a virtual method to `FakeMeshFixture_tmpl` that can return the desired coordinates, and then override that in a subclass? That way the initialisation will be done in the fixture setup and we don't need to add a class with what are effectively secondary setup/teardown methods.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+  return coords;
+}
+
+} // namespace
+
+TEST_F(DiamagneticDriftTest, CreateComponent) {
+  Options options = diamagneticOptions();
+  setCurvature();
   DiamagneticDrift component("test", options, nullptr);
 }
```
cmacmackin, line 95: We'll get this test for free once we merge #596, so we can probably leave it out here. (On which topic, I should probably check for similar tests in other components and remove them in PR #596.)

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+  Options state{
+      {"species",
+       {{"missing_temperature", {{"charge", 1.0}, {"density", 1.0}}},
+        {"missing_charge", {{"temperature", 1.0}, {"density", 1.0}}},
+        {"zero_charge", {{"charge", 0.0}, {"temperature", 1.0}, {"density", 1.0}}}}}};
+  Permissions permissions{{readOnly("species"),
+                           readWrite("species:missing_temperature:density_source"),
+                           readWrite("species:missing_charge:density_source"),
+                           readWrite("species:zero_charge:density_source")}};
+  GuardedOptions guarded_state{&state, &permissions};
```
cmacmackin, line 113: I strongly advise against creating `Permissions` and `GuardedOptions` objects when writing tests. It will increase fragility, as you risk the test breaking whenever permissions change on the `Component`. As explained earlier, we can write tests that only call the `DiamagneticDrift::transform` method, which will allow us to avoid using GuardedOptions here.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+
+  auto missing_temperature = guarded_state["species"]["missing_temperature"];
```
cmacmackin, line 115: What is the purpose of this line?

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+
+  Options state{
+      {"species",
+       {{"d+",
+         {{"charge", 1.0}, {"temperature", 1.0}, {"density", 1.0}, {"momentum", 1.0}}},
+        {"e", {{"charge", -1.0}, {"temperature", 1.0}, {"pressure", 1.0}}}}}};
+  Permissions permissions{{readOnly("species"), readWrite("species:d+:density_source"),
+                           readWrite("species:d+:momentum_source"),
+                           readWrite("species:e:energy_source")}};
+  GuardedOptions guarded_state{&state, &permissions};
```
cmacmackin, line 143: Again, don't create GuardedOptions objects. Use normal Options and call `DiamagneticDrift::transform`.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+
+  auto d_ion = guarded_state["species"]["d+"];
```
cmacmackin, line 145: What is the purpose of this line?

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+  ASSERT_TRUE(state["species"]["d+"].isSet("density_source"));
+  ASSERT_TRUE(state["species"]["d+"].isSet("momentum_source"));
+  ASSERT_FALSE(state["species"]["d+"].isSet("energy_source"));
+
+  ASSERT_FALSE(state["species"]["e"].isSet("density_source"));
+  ASSERT_TRUE(state["species"]["e"].isSet("energy_source"));
+  ASSERT_FALSE(state["species"]["e"].isSet("momentum_source"));
+}
+
+TEST_F(DiamagneticDriftTest, CoreAverageRequiresAverageCore) {
```
cmacmackin, line 159: If we make `coreAverage` a standalone function then we won't need this test anymore.

include/diamagnetic_drift.hxx
```diff
@@ -3,16 +3,51 @@
...
+
+  /// Calculate the diamagnetic sink for a single evolved quantity
+  /// using gradient form:
+  ///
+  ///   factor * Curlb_B . Grad(quantity * T / q)
+  Field3D calculateGradientForm(const Field3D& quantity, const Field3D& temperature,
+                                BoutReal charge, BoutReal factor = 1.0);
+
+  /// Add diamagnetic source terms for a single species subsection.
+  void addDiamagneticSources(GuardedOptions& species);
```
cmacmackin, line 37: It should not be necessary to make these public for unit testing. You can test them by passing simple states to `DiamagneticDrift::transform`. See, for example, the tests for the Braginskii closure. Making these methods public also increases the maintenance overhead for this class, as there is a chance that some other component could end up using them. This would make the class harder to refactor in future.

include/diamagnetic_drift.hxx
```diff
@@ -3,16 +3,51 @@
...
+
+  /// Smooth a source around the inner radial core boundary.
+  ///
+  /// Requires average_core to be enabled for this component.
+  /// Modifies the input field in-place.
+  void coreAverage(Field3D& f);
```
cmacmackin, line 43: Make this a standalone function (possibly in `bout_utils.hxx`) that takes additional arguments `const Field2D & cell_volume` and `BoutReal core_ring_volume`. That way it can be tested without being a public method on this class. It can also more easily be used elsewhere, if needed.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+    }
+  }
+
+  component.coreAverage(field);
+
+  const BoutReal expected_average = (1.0 * 1.0 + 2.0 * 2.0 + 3.0 * 4.0) / 6.0;
+  for (int jy = mesh->ystart; jy <= mesh->yend; ++jy) {
+    for (int jz = mesh->zstart; jz <= mesh->zend; ++jz) {
+      ASSERT_DOUBLE_EQ(field(mesh->xstart, jy, jz), expected_average);
+      ASSERT_DOUBLE_EQ(field(mesh->xstart + 1, jy, jz), 9.0);
```
cmacmackin, line 199: I'm not quite clear what this check is hoping to show. I'm guessing it's that the averaging wasn't applied except write next to the core. However, in that case it would be more effective to make this value vary with y and/or z, to demonstrate that it hasn't be averaged in any way.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+  div_mesh.xstart = 2;
+  div_mesh.xend = div_mesh.LocalNx - 3;
+  div_mesh.createDefaultRegions();
+  div_mesh.setCoordinates(nullptr);
+  auto div_coords = makeUnitCoordinates(&div_mesh);
+  div_mesh.setCoordinates(div_coords);
+  div_mesh.setGridDataSource(new FakeGridDataSource());
+  div_mesh.createBoundaryRegions();
+
+  ScopedGlobalMesh use_div_mesh(&div_mesh);
```
cmacmackin, line 215: As mentioned above, consider moving this logic into a fixture, to keep the body of the test cleaner.

src/diamagnetic_drift.cxx
```diff
@@ -1,36 +1,71 @@
...
 
-  bndry_flux =
-      options["bndry_flux"].doc("Allow fluxes through boundary?").withDefault<bool>(true);
-
-  diamag_form = options["diamag_form"]
-    .doc("Form of diamagnetic drift: 0 = gradient; 1 = divergence")
-    .withDefault(Field2D(1.0));
+  bndry_flux = options["bndry_flux"]
+                   .doc("Allow fluxes through boundary?")
+                   .withDefault<bool>(false);
```
cmacmackin, line 33: Seeing as we're making breaking changes anyway, we might as well rename this "boundary_flux".

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+
+  DiamagneticDrift component("diamagnetic_drift", options, nullptr);
+
+  Field3D quantity = makeField<Field3D>([](Ind3D& i) {
+    const bool on_y_boundary = i.y() <= mesh->ystart || i.y() >= mesh->yend;
+    return on_y_boundary ? 0.0 : 1.0 + 0.1 * i.z();
+  });
+  Field3D temperature{1.0};
+
+  Field3D sink = component.calculateDivergenceForm(quantity, temperature, 1.0);
```
cmacmackin, line 231: As described previously, I think it is better to do this with calls to `DiamagneticDirft::transform`. If you really want to do it this way, however, there is no need to set `average_core` or `diveregence_form`.

tests/unit/test_diamagnetic_drift.cxx
```diff
@@ -18,13 +21,234 @@ using namespace bout::globals;
...
+
+  ASSERT_NEAR(volumeIntegral(sink), 0.0, 1e-12);
+}
+
+TEST_F(WideDiamagneticDriftTest, GradientFormIsConservativeWithConstantBoundaryProduct) {
```
cmacmackin, line 236: What is meant by `BoundaryProduct` in this test name. I see the constant boundary, but where is the product?

docs/sphinx/equations.rst
```diff
@@ -946,33 +946,37 @@ equations. Calculates the diamagnetic drift velocity as
...
+equivalent away from boundaries, differing by the divergence of a curl.
 
-However, Form 1 is nice because the flow velocity depends on the temperature, not the pressure gradient. 
-This usually makes it better behaved numerically. To make the most of both, the `diamagnetic_drift` component allows the forms to be mixed
-using the ``diamag_form`` setting. For example, the :code:`tcv-x21` example blends it such that form 0 is at the boundary:
+At boundaries the distinction matters: the divergence form can correspond to
+a flux through the boundary, while the gradient form can be used to avoid
+those unphysical boundary fluxes.
 
+In Hermes-3 these are selected with the boolean option ``divergence_form``:
```
cmacmackin, line 959: Use an enum for this, instead of a boolean.

cmacmackin (changes requested):
I don't like the way public functions have been introduced in this. They are not necessary to test the behaviour of the component and will increase code maintenance burdens.

Even if you decide to keep `calculateDivergenceForm` and `calculateGradientForm` public, I would strongly urge you to make `addDiamagneticDrift` private. I see no value in calling it directly in your unit tests and doing so requires setting up `GuardedOptions` objects. Doing that will require you to keep track of the permissions set by `DiamagneticDrift` and make your tests significantly more fragile as a result..

PR #541, opened 2026-04-09
Fix diffusion-transport example by not reading density boundary

cmacmackin:
Fixes #536.


mikekryjak (changes requested):
Thanks @cmacmackin. Only one suggestion - the collision frequencies are currently also reading the whole domain, but they are only meaningful on cell centres, where they are used to calculate the time derivatives. Because of this, we should probably use `GET_NOBOUNDARY` and restrict their permissions to interior only as well.

bendudson:
> Thanks @cmacmackin. Only one suggestion - the collision frequencies are currently also reading the whole domain, but they are only meaningful on cell centres, where they are used to calculate the time derivatives. Because of this, we should probably use `GET_NOBOUNDARY` and restrict their permissions to interior only as well.

This is now also changed.

PR #542, opened 2026-04-10
Add neutral equation term diagnostics

sun51027:
Add neutral equations physical term diagnostics.
Just a draft, haven't completed yet.


mikekryjak:
This will be redone on top of the new limiter branch once it's done.

PR #543, opened 2026-04-15
Neutral passive momentum

Vandoo:
Add an option to solve only the neutral density equation, with ( Td = Td+ ) as described in https://doi.org/10.1016/j.nme.2024.101824 (the definition of D_n remains unchanged). The parallel mometum is passively given by the local reaction balance (Nn_eq) * U(the ion parallel velocity), assuming a highly collisional regime. 

Initial tests in an early transition phase indicate that, for the same simulated physical time, this approach could reduce computational cost by ~60%.

### DIFF
--- a/src/neutral_mixed.cxx
+++ b/src/neutral_mixed.cxx
@@ -41,9 +41,20 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
   evolve_momentum = options["evolve_momentum"]
                         .doc("Evolve parallel neutral momentum?")
                         .withDefault<bool>(true);
+  passive_momentum = options["passive_momentum"]
+                        .doc("Evolve only neutral density, passive momentum+pressure")
+                        .withDefault<bool>(false);
+  temperature_from = options["temperature_from"]
+                .doc("Name of species to take temperature from. If not set, uses own species temperature.")
+                .withDefault<std::string>("");
 
   if (evolve_momentum) {
     solver->add(NVn, std::string("NV") + name);
+  } else if (passive_momentum) (
+        "WARNING: Not evolving neutral parallel momentum. "
+        "NVn set from diffusion + ion flow.\n");
+    NVn = 0.0;
+    Vn = 0.0;
   } else {
     output_warn.write(
         "WARNING: Not evolving neutral parallel momentum. NVn and Vn set to zero\n");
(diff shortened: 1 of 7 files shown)


src/neutral_mixed.cxx
```diff
@@ -40,21 +42,28 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
   evolve_momentum = options["evolve_momentum"]
                         .doc("Evolve parallel neutral momentum?")
                         .withDefault<bool>(true);
+  passive_momentum = options["passive_momentum"]
```
mikekryjak, line 45: @Vandoo Thanks for this PR, I'd like to try it as well in my SOLPS comparison if I have the time. I have a few questions. First one - is there some reason why you lumped in the passive momentum and Tn = Ti together into one option? is it possible to separate them?

src/neutral_mixed.cxx
```diff
@@ -506,111 +482,164 @@ void NeutralMixed::finally(const Options& state) {
...
+    }
+    if (!equilibrium_nu_rec_name.empty()) {
+      nu_rec = GET_VALUE(Field3D, localstate["collision_frequencies"][equilibrium_nu_rec_name]);
     }
-    ddt(NVn) += Snv;
 
+    // Equilibrium neutral density:
+    // Nn_eq = Ni * (Nn * nu_cx + Ne * nu_rec) / (Ni * nu_cx + Ne * nu_iz)
+    // Quasineutrality: Ni = Ne
+    Field3D Nn_eq = (Nnlim * nu_cx + Ne * nu_rec)
```
mikekryjak, line 632: How come you can't use the existing collisionality block? I wrote this collisionality thing, it's really messy and seeing it twice in one file makes it even worse! I need to refactor it when I have the time, but in the meantime maybe the first block can be reused?

src/neutral_mixed.cxx
```diff
@@ -321,98 +295,103 @@ void NeutralMixed::finally(const Options& state) {
...
+    Tn = GET_VALUE(Field3D, localstate["temperature"]);
+    Pn = Nn * Tn;
+    Pnlim = softFloor(Pn, pressure_floor);
+    Pnlim.applyBoundary();
+  }
+
   logPnlim = log(Pnlim);
   logPnlim.applyBoundary();
-  ///////////////////////////////////////////////////////
   // Calculate cross-field diffusion from collision frequency
```
dschwoerer, line 310:
```suggestion
  logPnlim.applyBoundary();

  ///////////////////////////////////////////////////////
  // Calculate cross-field diffusion from collision frequency
````

dschwoerer (changes requested):
There is also some commented out code, that should be removed.

bendudson:
Hi @Vandoo ! Are you not planning to merge this? It looked like a promising approach. There are lots of changes being made in the neutral model, but it would be great to have this option.

Vandoo:
Hello @bendudson , thanks for asking. Sorry, I was setting up a repo on my own account and accidentally deleted the branch from master (which closed the old PR). The old branch had drifted well behind master, and I'm still testing this (the
  extra per-RHS cost isn't cheap). I'll re-implement the option on current master and open a fresh PR.

PR #548, opened 2026-04-20
Do not compute invalid parallel slices

dschwoerer:


totork:
This will probably break some if not most simulations because of the way the old BOUT++ version works there. It is briefly explained in 
https://bout.zulipchat.com/#narrow/channel/535664-BSTING/topic/zoidberg.20J_forward.2Fbackward.20issue/with/604220321

dschwoerer:
Sorry, no. This does not explain it. Did you post the correct link?

Do the parallel slices of J get used somewhere? If so, that is a bug, that better throws an exception then silently do the wrong thing.
If it is not, then it is still better to not have the wrong data stored, to avoid introducing regressions that use it.

If this change breaks anything, it was broken before, and it is better to notice, then to have wrong results. After all, any uses should be fixable, but only if we know.

dschwoerer:
@totork I think it should be good now

totork:
LGTM

totork (approved):

PR #549, opened 2026-04-20
Update BOUT++ to latest version and fix sheath_ion_heat_flux

dschwoerer:


dschwoerer:
[a826725](https://github.com/boutproject/hermes-3/pull/549/commits/a8267255dc062f70edbb84abf3b9858b64f5fb95) might be impacting some simulations, depending on the normalisation.

mikekryjak:
What was the issue with `sheath_ion_heat_flux` and what is the fix?

dschwoerer:
from https://github.com/boutproject/hermes-3/pull/549/changes/a8267255dc062f70edbb84abf3b9858b64f5fb95:

> Do not convert to int
> 
> abs is for integers (from C)
> std::abs is for anything (from C++)

Previously, the flux was converted to an integer, and thus was reduced. How much, depends on the normalisation.

mikekryjak:
No way.. I didn't know about that. Have you checked if there are other instances of this elsewhere in the code?

dschwoerer:
I did not see any other compiler warning, so I think it is fine ...

PR #550, opened 2026-04-20
[WIP] Add metric normalisation for Fci, throw error if recalcuated while Fci.

totork:


dschwoerer:
I pushed a fixed version of your commit here:
https://github.com/boutproject/hermes-3/tree/master-fci-normalisation

This now also applies to Hermes:
https://bout.zulipchat.com/#narrow/channel/504946-BOUT.2B.2B/topic/Change.20to.20dev.20workflow.20and.20CI/with/580633616

PR #551, opened 2026-04-21
Add metric normalisation for Fci, throw error if recalcuated while Fci.

dschwoerer:
Formatted version of #550


include/neutral_full_velocity.hxx
```diff
@@ -100,5 +102,5 @@ namespace {
 RegisterComponent<NeutralFullVelocity>
     registersolverneutralfullvelocity("neutral_full_velocity");
 }
-
+#endif // fu
```
ZedThree, line 105:
```suggestion
#endif // not BOUT_USE_METRIC_3D
```

src/div_ops.cxx
```diff
@@ -965,919 +981,922 @@ const Field3D Div_a_Grad_perp_upwind(const Field3D& a, const Field3D& f) {
...
+        flow_ylow.setDirectionY(YDirectionType::Aligned);
+      }
 
-  for (int i = mesh->xstart; i <= mesh->xend; i++) {
-    for (int j = mesh->ystart; j <= mesh->yend; j++) {
-      for (int k = 0; k < mesh->LocalNz; k++) {
-        // Calculate flux between j and j+1
-        int kp = (k + 1) % mesh->LocalNz;
-        int km = (k - 1 + mesh->LocalNz) % mesh->LocalNz;
+      if (bout::build::use_metric_3d) {
```
ZedThree, line 1180:
```suggestion
      if constexpr (bout::build::use_metric_3d) {
```

src/div_ops.cxx
```diff
@@ -965,919 +981,922 @@ const Field3D Div_a_Grad_perp_upwind(const Field3D& a, const Field3D& f) {
...
+    ///           +---|---+
+    ///           |       |
+    /// xlow(i,j, k) -> (i,j, k) -> xlow(i+1,j, k)
+    ///           |   ^   |
+    ///           +---|---+
+    ///           ylow(i,j, k)
+    ///
+    ///
+    /// WARNING: Causes checkerboarding in neutral_mixed integrated test
+    const Field3D Div_a_Grad_perp_upwind_flows(const Field3D& a, const Field3D& f,
```
ZedThree, line 1477: Probably needs fixing on the header declaration too

```suggestion
    Field3D Div_a_Grad_perp_upwind_flows(const Field3D& a, const Field3D& f,
```

src/div_ops.cxx
```diff
@@ -965,919 +981,922 @@ const Field3D Div_a_Grad_perp_upwind(const Field3D& a, const Field3D& f) {
...
-        BoutReal gradient =
-            // df/dz
-            (fc(i, j, kp) - fc(i, j, k)) / coord->dz(i, j)
+    for (int i = mesh->xstart; i <= mesh->xend; i++) {
+      for (int j = mesh->ystart; j <= mesh->yend; j++) {
+#if BOUT_USE_METRIC_3D
+        for (int k = 0; k < mesh->LocalNz; k++) {
+#else
+    int k = 0;
+#endif
```
ZedThree, line 1052: These constructs make me itchy, I really don't like splitting up statements with preprocessor guards like this, it makes the code quite fragile.

Instead, we can do:

```cpp
    const int zend = bout::build::use_metric_3d ? mesh->LocalNz - 1 : 0;

    for (int i = mesh->xstart; i <= mesh->xend; i++) {
      for (int j = mesh->ystart; j <= mesh->yend; j++) {
        for (int k = 0; k <= zend; k++) {
```

For 2D metrics, the compiler should optimise out the loop completely.

We might also want to make these loops ready for Z-guards:

```cpp
    const auto zstart = bout::build::use_metric_3d ? localmesh->zstart : 0;
    const auto zend = bout::build::use_metric_3d ? localmesh->zend : 0;

    for (int i = mesh->xstart; i <= mesh->xend; i++) {
      for (int j = mesh->ystart; j <= mesh->yend; j++) {
        for (int k = zstart; k <= zend; k++) {
```

include/neutral_full_velocity.hxx
```diff
@@ -9,6 +9,8 @@
 
 #include <bout/vector2d.hxx>
 
+#if not BOUT_USE_METRIC_3D
```
ZedThree, line 12: Do we need something like `RegisterUnavailableComponent` to be able to give a useful error message about the incompatibility?

ZedThree (changes requested):
Looks good to me, but I have a few suggestions

src/div_ops.cxx
dschwoerer: And then a second for loop inside, with:
```
          const auto zstart = bout::build::use_metric_3d ? k : localmesh->zstart;
          const auto zend = bout::build::use_metric_3d ? k : localmesh->zend;
          for (int k = zstart; k <= zend; k++) {
```
?

dschwoerer:
Thanks for the review @ZedThree, I have applied most of the comments in the #577 PR.

I am not sure about RegisterUnavailableComponent - I think as long as it just one component for FCI metrics, it is probably not worth it. We can revisit it, if we have more components like this.

Issue #547, opened 2026-04-17
Failed to run example case 1D-neon-source


lqs661:
When I tried to run the 1D-neon-source case in example/tokamak-1D/extra, I met some problems. The log indicated the following error:
#0  (filtered)
#1  (filtered)
#2  in Options& set<Field3D>(Options&, Field3D) [clone .isra.0] [clone .cold]
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/hermes-3
#3  in decltype(auto) add<Field3D, GuardedOptions, void>(GuardedOptions&&, Field3D)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/hermes-3
#4  in NeutralParallelDiffusion::transform_impl(GuardedOptions&)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/libhermes-3-lib.so
#5  in Component::transform(Options&)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/libhermes-3-lib.so
#6  in ComponentScheduler::transform(Options&)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/libhermes-3-lib.so
#7  in Hermes::rhs(double, bool)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/hermes-3
#8  in Solver::run_rhs(double, bool)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/external/BOUT-dev/lib/libbout++.so.5.2.0
#9  in SNESSolver::rhs_function(_p_Vec*, _p_Vec*, bool)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/external/BOUT-dev/lib/libbout++.so.5.2.0
#10 in SNESSolver::snes_function(_p_Vec*, _p_Vec*, bool)
    at /public3/home/t6s010954/hermes-dev/hermes-3-boron-test/build/external/BOUT-dev/lib/libbout++.so.5.2.0
#11 in SNESComputeFunction
    at /public3/home/t6s010954/library_mine/petsc/src/snes/interface/snes.c:2496:7
    2494:       SNESFunctionFn *computefunction;
    2495:       PetscCall(DMSNESGetFunction(dm, &computefunction, &ctx));
  > 2496:       PetscCallBack("SNES callback function", (*computefunction)(snes, x, y, ctx));
                ^
    2497:     }
[...]

PR #553, opened 2026-04-23
Fixes for 1D examples

bendudson:
Going through examples, to fix issue #547 

- tokamak-1D/extra/1D-hydrogen
- tokamak-1D/extra/1D-neon
- tokamak-1D/extra/1D-neon-source

### DIFF
--- a/src/braginskii_collisions.cxx
+++ b/src/braginskii_collisions.cxx
@@ -150,7 +150,7 @@ void BraginskiiCollisions::collide(GuardedOptions& species1, GuardedOptions& spe
     });
 
     add(species2["collision_frequency"], nu); // Total collision frequency
-    std::string const coll_name =
+    const std::string coll_name =
         species2.name() + std::string("_") + species1.name() + std::string("_coll");
     set(species2["collision_frequencies"][coll_name],
         nu); // Collision frequency for individual reaction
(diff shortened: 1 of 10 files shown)


src/braginskii_collisions.cxx
```diff
@@ -370,7 +372,7 @@ void BraginskiiCollisions::transform_impl(GuardedOptions& state) {
             BoutReal const coulomb_log =
                 29.91
                 - log((Z1 * Z2 * (AA1 + AA2)) / (AA1 * Tlim2 + AA2 * Tlim1)
-                      * sqrt(Nlim1 * SQ(Z1) / Tlim1 + Nlim2 * SQ(Z2) / Tlim2));
+                      * sqrt((Nlim1 * SQ(Z1) / Tlim1) + (Nlim2 * SQ(Z2) / Tlim2)));
```
mikekryjak, line 375: I am hoping this change is just style preference and C++ doesn't do something bizarre to make these brackets necessary?

src/braginskii_collisions.cxx
```diff
@@ -390,6 +392,10 @@ void BraginskiiCollisions::transform_impl(GuardedOptions& state) {
         } else {
           // species1 charged, species2 neutral
 
+          if (!ion_neutral) {
```
mikekryjak, line 395: Thanks for spotting this. It could explain the difference we saw in the ReMKiT1D benchmark when adding neutrals!

mikekryjak:
Thanks for this @bendudson. I've gone through your changes and I can see that there must have been an issue where `nu` was zero, but how did this happen? Surely `braginskii_collisions` has several collisions on by default, including the `ion_neutral` one which as you found was always on.

src/braginskii_collisions.cxx
dschwoerer: That is clang-tidy - it likes to make it obvious that `*` has precedence before `+` :shrug:

bendudson:
Hey @mikekryjak ! The `neutral_parallel_diffusion` uses the "afn" collisions by default, which just include ionisation and charge exchange, not neutral-neutral collisions. The ion density went to zero at the target so the afn collision frequency went to zero. In the examples I've switched it to "multispecies" so that neutral-neutral collisions are also included.

PR #554, opened 2026-04-24
External Apar flutter

bendudson:
Allows an external perturbation to be added to `Apar_flutter`


include/external_apar.hxx
```diff
@@ -0,0 +1,35 @@
+#pragma once
+#ifndef EXTERNAL_APAR_H
+#define EXTERNAL_APAR_H
```
cmacmackin, line 3: Do we need these guards if we are also using `#pragma once`

src/external_apar.cxx
```diff
@@ -0,0 +1,58 @@
...
+  // Normalisations
+  auto Bnorm = get<BoutReal>(state["Bnorm"]);
+  auto rho_s0 = get<BoutReal>(state["rho_s0"]);
+
+  set_with_attrs(state["external_apar"], external_apar,
+                 {{"units", "Tm"},
+                  {"conversion", Bnorm * rho_s0},
+                  {"standard_name", "External Apar"},
+                  {"long_name", "External Apar"},
+                  {"source", "external_apar"}});
```
cmacmackin, line 57: Is this something we'll always want to save to the output file? It might be better to add a `diagnose` parameter to the component and only write this out if `diagnose == true`.

tests/unit/test_external_apar.cxx
```diff
@@ -0,0 +1,137 @@
...
+
+/// Global mesh
+namespace bout {
+namespace globals {
+extern Mesh* mesh;
+} // namespace globals
+} // namespace bout
+
+// The unit tests use the global mesh
+using namespace bout::globals;
```
cmacmackin: Is there a reason you have to use this namespace? References to `bout::globals::mesh` all appear to be qualified anyway.

tests/unit/test_external_apar.cxx
```diff
@@ -0,0 +1,137 @@
...
+                     {"test", {{"scale", 4.0}}}};
+
+  const BoutReal external_apar = 3.1;
+  dynamic_cast<FakeMesh*>(bout::globals::mesh)
+      ->setGridDataSource(new FakeGridDataSource{{{"external_apar", external_apar}}});
+
+  ExternalApar component("test", options, nullptr);
+
+  Options state{{"fields", {{"Apar_flutter", 2.0}}}};
+  component.transform(state);
```
cmacmackin, line 121: Do we actually need to call `transform` for this test?

cmacmackin (approved):
Looks good, just a few minor questions about how you've done things.

include/external_apar.hxx
bendudson: This used to be needed for older/proprietary compilers so I still do it out of habit. Probably not necessary but I don't think it does any harm.

src/external_apar.cxx
bendudson: Probably yes: It's only saved once because it's not time dependent, so it doesn't take much space relatively speaking.

tests/unit/test_external_apar.cxx
bendudson: Removed, thanks!

tests/unit/test_external_apar.cxx
bendudson: I wanted to check that transform didn't modify the output somehow (e.g. modifying the internal state). I've split into two tests: One calls `transform` and the other doesn't.

PR #571, opened 2026-05-22
Fci: Add core dissipation of vorticity

totork:
This PR adds a way to dissipate vorticity in the core. The problem is, that the core is a source of instabilities which slow down convergence. Dissipating all the vorticity in the core leads to exact force balance.


bshanahan:
LGTM, Thanks @totork!

PR #572, opened 2026-05-22
Fci: Add zonal neumann BC for potential

totork:
This PR adds an additional solve for the potential, that takes the a z-averaged potential as the inner boundary condition to have no radial ExB fluxes in/out of the core. 

This PR needs the merge of the 3D MMS tests


bshanahan:
Is this ready to go once we merge #537?

totork:
I can not reproduce the failing inversion test, for me they pass. Otherwise it should be fine.

PR #576, opened 2026-06-09
Add ionising core feature

sun51027:
Principle of the “ionising core” feature: Neutrals that reach the core boundary are removed from the domain and their flux is added to the boundary condition on the flux of ions from the core. In other words, neutrals are ionised when hit the core boundary and turn back as ions.

It aims to make the neutral core boundary treatment more physical, as well as to replicate the core boundary feature used in SOLPS.

New features are added:
1. `ionising_core=true`: remove neutral particles/momentum/energy flow toward the core boundary, while neutral particle flow return as ion particle flow.
2. `ionise_as`: ion species, e.g. `d` -> `d+`, `t` -> `t+`
3. `ionising_core_return_mom_energy=true`: neutral momentum/energy flow will return as ion momentum/energy flow
4. `ionising_core_iz_energy_loss`: User defined variable for electron energy pay for ionisation. The default value is 13.6 (eV). Can be set to 0 to disable this feature.

**WIP: Ionisation at core: neutral goes into core, coming back as ion**

---

**To-do**
- [x] Renamed module from `core_ionising` → `ionising_core`

- [x] Add flag `ionisation_energy_loss`
  - If `true`: cost energy from electrons to ionise neutral
  - Formula: `electron energy source -= ionise_particle_flow / vol`
  - [x] Rename: `ionisation_energy_loss` → `ionising_core_iz_energy_loss`
  - [x] Set as a user-defined variable for user to choose the energy cost for ionisation.
 

- [x] Add flag `only_particle_flow` (SOLPS-like) (
  - Aims to reproduce SOLPS's feature, which only account for paritcle flux, not energy/momentum flux
  - [x]  Rename: only_particle_flow → ionising_core_return_mom_energy

- [x] Use parallel momentum instead of one-way momentum flux (radial momentum)

- [ ] Write and pass regression / unit tests
- [ ] Documentation


mikekryjak:
I see that this PR contains this one: https://github.com/boutproject/hermes-3/pull/451
Can you change this PR to reference that branch instead of the master branch for now? This will make it easier to see changes. Right now, because this PR contains both branches and neither is merged, the diff shows changes from both changes at the same time.

Once I merge #556, we can point this PR at master.

Issue #573, opened 2026-05-31
Compilation error when using 3D metrics


totork:
When I want to compile the master branch with 

-DBOUT_ENABLE_METRIC_3D=ON

I do get the error

`
/u/toto/soft/boutproject/hermes-3/src/braginskii_ion_viscosity.cxx: In constructor ‘BraginskiiIonViscosity::BraginskiiIonViscosity(const std::string&, Options&, Solver*)’:
/u/toto/soft/boutproject/hermes-3/src/braginskii_ion_viscosity.cxx:120:13: error: no match for ‘operator*=’ (operand types are ‘Vector2D’ and ‘Field3D’)
  120 |     Curlb_B *= 2. / coord->Bxy;
      |     ~~~~~~~~^~~~~~~~~~~~~~~~~~
In file included from /u/toto/soft/boutproject/hermes-3/external/BOUT-dev/include/bout/fv_ops.hxx:12,
                 from /u/toto/soft/boutproject/hermes-3/src/braginskii_ion_viscosity.cxx:13:
/u/toto/soft/boutproject/hermes-3/external/BOUT-dev/include/bout/vector2d.hxx:104:13: note: candidate: ‘Vector2D& Vector2D::operator*=(BoutReal)’
  104 |   Vector2D& operator*=(BoutReal rhs);
      |             ^~~~~~~~
/u/toto/soft/boutproject/hermes-3/external/BOUT-dev/include/bout/vector2d.hxx:104:33: note:   no known conversion for argument 1 from ‘Field3D’ to ‘BoutReal’ {aka ‘double’}
  104 |   Vector2D& operator*=(BoutReal rhs);
      |                        ~~~~~~~~~^~~
/u/toto/soft/boutproject/hermes-3/external/BOUT-dev/include/bout/vector2d.hxx:107:13: note: candidate: ‘Vector2D& Vector2D::operator*=(const Field2D&)’
  107 |   Vector2D& operator*=(const Field2D& rhs);
      |             ^~~~~~~~
/u/toto/soft/boutproject/hermes-3/external/BOUT-dev/include/bout/vector2d.hxx:107:39: note:   no known conversion for argument 1 from ‘Field3D’ to ‘const Field2D&’
  107 |   Vector2D& operator*=(const Field2D& rhs);
      |                        ~~~~~~~~~~~~~~~^~~
[...]

PR #577, opened 2026-06-10
div_ops: Compile with METRIC_3D

bendudson:
Modify Div_par_fvv and Div_par_fvv_heating so that they compile with 3D metrics.

Maybe helps address issue #573 

Ready to merge. The patch coverage is below the threshold, but code coverage increases with this PR because a lot of uncovered code is removed (or rather, moved to BOUT-dev).


bshanahan (commented):
Looks OK to me -- don't think the test failures are related to this PR.

dschwoerer (approved):

dschwoerer:
The testing coverage is not really helpful, as so far 3D metrics only enables compiling. There will be more tests coming.

ZedThree (approved):
LGTM

src/neutral_mixed.cxx
```diff
@@ -118,6 +118,9 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
                            .doc("Include neutral gas heat conduction?")
                            .withDefault<bool>(true);
 
+  conduction_method =
+      options["conduction_method"].withDefault<std::string>(conduction_method);
```
ZedThree, line 122: This is possibly now out of sync with BOUT++ and needs to be an enum, but maybe this PR can go in first and then that's a simple update?

hermes-3.cxx
```diff
@@ -121,9 +121,9 @@ class DecayLengthBoundary : public BoundaryOp {
 
     // Get cell radial length
     Coordinates* coord = mesh->getCoordinates();
-    Field2D dx = coord->dx;
-    Field2D g11 = coord->g11;
-    Field2D dr =
+    auto dx = coord->dx;
+    auto g11 = coord->g11;
+    Coordinates::FieldMetric dr =
```
ZedThree, line 126: Is it now true that this can't be `auto` otherwise it becomes an expression template?

ZedThree (approved):
LGTM! Just tons of small but obvious fixes for 3D metrics

src/neutral_mixed.cxx
dschwoerer: Indeed, I would need to update BOUT++ to the final version that got merged, and then update here.
I will make a PR into this one, then you can decide on the order of merging.

hermes-3.cxx
dschwoerer: Indeed.

src/neutral_mixed.cxx
dschwoerer: With https://github.com/boutproject/BOUT-dev/pull/3425 merged, it probably makes sense to first merge this, and then later update to BOUT++ and update this too.

PR #579, opened 2026-06-12
Fix tests in master

bendudson:
Changes to xhermes broke integrated test scripts


bendudson:
Hi @mikekryjak Would you mind having a look at this? It looks like the region selection changes in `xhermes` have broken the `2D-recycling` test.

PR #582, opened 2026-06-16
Automatic PRs to update xhermes/boutdata/xbout

mikekryjak:
The dependencies are installed from a txt list, and there is a bot to update the list at 6am every morning.
Not done this before, so any comments/criticism welcome!

Instead of using dependabot, this uses a simpler approach through https://github.com/peter-evans/create-pull-request.


dschwoerer:
Why does codecov think this reduces coverage?

dschwoerer (approved):
Looking forward to testing this in action!

ZedThree:
If we put these dependencies in the `pyproject.toml`, could `dependabot` just keep them up-to-date for us?

mikekryjak:
I don't have experience with dependabot. My stochastic parrot says that it would be possible with sufficient config, but it wouldn't be as clean. 

In this system, there is only one PR updating xBOUT, xHermes and boutdata in one. You have more experience in this than me, so if you think dependabot is better, we could have a go... but maybe we can merge this in the meantime so that we have something in place.

PR #587, opened 2026-06-25
Fix tests again

mikekryjak:
This commit (https://github.com/boutproject/hermes-3/commit/c1f59154784fa6c9d62766877aebecaab43b1c05) introduced a workaround for the new xHermes selectors which blended the new selector system with old style index selection. This was a narrow fix for the test but broke it when `gen_data=True`. 

I also spent a bit of time confusing myself about why `2D-recycling` omits edge domain cells, thinking something could be wrong. I update comments/docs to make it clearer for the next person. Prek made formatting changes to the rest of the file as well.


tests/integrated/2D-production/runtest
oparry-ukaea, line 194: Could we also add an `exit(2)` after this for loop to guard against accidentally committing the file with gen_data=True.

mikekryjak: Good catch, thank you - just done

bendudson (approved):
Thanks @mikekryjak ! Ignoring the Codecov failure as this only changes integrated test scripts and docs.

PR #591, opened 2026-06-26
setBoundaryTo - revert to legacy version

### LINKED ISSUE #588: Recent changes change component order requirements for braginskii_conduction and sheath_boundary_simple
## Problem
Installing Hermes-3 from Spack and running MAST-U simulations with this component order:

```
components = (d+, d, e, c,  braginskii_collisions, 
                braginskii_friction, braginskii_heat_exchange,braginskii_conduction,
                sheath_boundary_simple, recycling, sound_speed, reactions, electron_force_balance)
```

Results in this not very verbose message:

```
Setting boundary of density but it has already been used in .
```

## Analysis
As far as I can tell, this originated from https://github.com/boutproject/BOUT-dev/pull/3308 where `setBoundaryTo()` was removed. @dschwoerer  What was the reason for it being removed? The PR message doesn't mention it.

The Hermes-3 commit removing all the `setBoundaryTo()` is here: https://github.com/boutproject/hermes-3/commit/870c800f5d037d89959cca7b15a47e81a9a30011 as part of https://github.com/boutproject/hermes-3/pull/578. 

My LLM says that this causes a component ordering change. Now that `braginskii_conduction` gets the full fields of `N`, `T` and `P` instead of just resetting their boundaries, downstream components can no longer set them, which means that `sheath_boundary_simple` cannot apply the sheath boundary!

I have tested this and found that indeed the bug is resolved if you swap `braginskii_conduction` and `sheath_boundary_simple`. 

According to my LLM, there is no verbosity in the exception because of the use of
```NVn = get<Field3D>(localstate["momentum"]);``` 
instead of:
```NVn = GET_VALUE(localstate["momentum"]);```
Is this correct? I'm not familiar with how this system works. 

## Solution?
I'm not an expert on component ordering, but don't these changes constrain it a lot? If we have to mark the entire field as immutable every time we apply boundaries, then surely that could cause some issues... My LLM is suggesting that `braginskii_conduction` doesn't need boundaries anyway so you could just read the domain and avoid the issue, but this still leaves us more constrained in general. 

Also, it would be really really good if we could have verbosity for these issues to debug component order problems.

dschwoerer:
Does this resolve https://github.com/boutproject/hermes-3/issues/588 ?


dschwoerer:
I assume we do not want the legacy behaviour, but do we want to keep using `setBoundaryTo`?
If not, should it be deprecated in BOUT++, if Hermes does not use it anymore?

PR #592, opened 2026-06-26
Add passive_momentum option to neutral_mixed

Vandoo:
Re-applies the previously-proposed neutral_passive_momentum feature onto current master (the original branch had drifted ~600 commits behind).

New options:
- passive_momentum: evolve only the neutral density; momentum and pressure are passive (pressure is not added to the solver).
- temperature_from: take Tn from another species (e.g. the ion, Tn=Ti).
- evolve_pressure: defaults to !passive_momentum.
- anomalous_transport_momentum (AnomalousDiffusion): toggle anomalous transport of parallel momentum. (not nessesary).

In passive_momentum mode NVn is set from an ion-equilibrium relation NVn = Nn_eq * Vi, with Nn_eq built from cx/iz/rec collision frequencies, and nu_cx/nu_iz/nu_rec/Nn_eq exposed as diagnostics. 

The preconditioner gains a density-only branch (I - gamma*Div_perp(Dnn*Grad_perp))*ddt(Nn) when pressure is not evolved. Pressure/momentum terms, sources, diagnostics, time-derivative scaling, freeze_low_density and CHECK blocks are gated on the new flags. Reading another species' temperature declares the required GuardedOptions read permission.

Tested on a 3D MAST-U CD case: builds clean, runs to completion with finite fields, reproduces the original branch's passive behaviour (Tn tracks Ti to ~0.1%, 7 evolved fields, consistent neutral diagnostics).


src/neutral_mixed.cxx
```diff
@@ -35,26 +35,56 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
...
   // Need to take derivatives in X for cross-field diffusion terms
   ASSERT0(mesh->xstart > 0);
 
   auto& options = alloptions[name];
 
-  // Evolving variables e.g name is "h" or "h+"
-  solver->add(Nn, std::string("N") + name);
-  solver->add(Pn, std::string("P") + name);
-
   evolve_momentum = options["evolve_momentum"]
```
mikekryjak, line 56: It looks like when `passive_momentum` is enabled, you need to disable `evolve_momentum` manually or else it'll be evolved. 

One solution could be to make `evolve_momentum` default to `!passive_momentum`, and raise a `BoutException` when both are enabled.

src/neutral_mixed.cxx
```diff
@@ -329,6 +366,14 @@ void NeutralMixed::finally(const Options& state) {
   // Field3D logNn = log(Nn);
   // Field3D logTn = log(Tn);
 
+  // In passive_momentum mode the pressure is not evolved; set_temperature has
+  // run by now, so localstate["temperature"] holds the current Ti. Recompute
+  // Pn = Nn*Tn so the derived (Dnn, kappa_n, logPnlim, Pnlim) stay consistent.
+  if (!evolve_pressure && !temperature_from.empty()) {
+    Tn = get<Field3D>(localstate["temperature"]);
+    Pn = Nn * Tn;
+  }
```
mikekryjak, line 391: What happens if `evolve_pressure = False` and `temperature_from = ""`?

src/neutral_mixed.cxx
mikekryjak, line 570: The density equation needs `Vn`. But at this point in the code, `Vn` is taken from the state, instead of the later recalculation from the ion velocity! This makes it inconsistent. 

Usually both `Vn` and `Nn` are evolved by the solver in `transform()`. This means both the density and momentum equations read a consistent, immutable state to prepare time derivatives for the next iteration. In this normal case, it's OK to have the momentum equation after density. 

However, now you are recalculating `Vn` in `finally()` from the ion velocity. This means that now `finally()` is inconsistent, with the density equation getting the "previous iteration". 

If I'm right, this would still run but probably cause performance issues. I think you need to move the velocity calculation to the top.

src/neutral_mixed.cxx
```diff
@@ -611,14 +669,64 @@ void NeutralMixed::finally(const Options& state) {
...
+        if (collisionSpeciesMatch(coll_name, std::string(name) + "+", name, "rec",
+                                  "partial")) {
+          nu_rec = GET_VALUE(Field3D, localstate["collision_frequencies"][coll_name]);
+          break;
+        }
+      }
+    }
+
+    Field3D Nn_eq = (Nnlim * nu_cx + Ne * nu_rec) / softFloor(nu_cx + nu_iz, density_floor);
+    NVn = Nn_eq * U;
```
mikekryjak: This is missing AA, so is off by a factor of 2 for `d`.

src/neutral_mixed.cxx
```diff
@@ -611,14 +669,64 @@ void NeutralMixed::finally(const Options& state) {
...
+        const std::string coll_name = collision.second.name();
+        if (collisionSpeciesMatch(coll_name, std::string(name) + "+", name, "rec",
+                                  "partial")) {
+          nu_rec = GET_VALUE(Field3D, localstate["collision_frequencies"][coll_name]);
+          break;
+        }
+      }
+    }
+
+    Field3D Nn_eq = (Nnlim * nu_cx + Ne * nu_rec) / softFloor(nu_cx + nu_iz, density_floor);
```
mikekryjak: There is a bug here, it floors collisionality with `density_floor`. The correct floor to use here is `Rnn`, which is derived from the "maximum neutral mean free path" `neutral_lmax`.

src/anomalous_diffusion.cxx
```diff
@@ -50,6 +50,10 @@ AnomalousDiffusion::AnomalousDiffusion(std::string name, Options& alloptions, So
                               .doc("Allow anomalous diffusion into sheath?")
                               .withDefault<bool>(false);
 
+  anomalous_transport_momentum = options["anomalous_transport_momentum"]
```
mikekryjak, line 55: What is this used for? The anomalous transport of momentum is just the viscosity `eta`. To disable it, just set `eta` to zero.

src/neutral_mixed.cxx
```diff
@@ -611,14 +669,64 @@ void NeutralMixed::finally(const Options& state) {
...
     }
     Snv = momentum_source;
     if (localstate.isSet("momentum_source")) {
       Snv += get<Field3D>(localstate["momentum_source"]);
     }
     ddt(NVn) += Snv;
 
+  } else if (passive_momentum) {
+    // NVn set from an equilibrium flow with the ions (not evolved):
+    //   NVn = Nn_eq * Vi,  Nn_eq = (Nn*nu_cx + Ne*nu_rec) / (nu_cx + nu_iz)
```
mikekryjak: I haven't quite understood how this works, and why you aren't using plain `Nn`. Let's discuss this offline. It would be good to explain it in documentation. If these collisionalities are really needed, then I would ask that they are encapsulated in a function defined somewhere in the top of the file so they don't break up the flow so much. Maybe you could do this for the other collisionality block as well at that point. I can help with this if you need!

This is really my fault, I wrote this collisionality system and it's super ugly, so sorry... I will refactor it at some point!

src/neutral_mixed.cxx
mikekryjak: I would also probably floor `nu_cx` and `nu_rec` to `Rnn` in the numerator for consistency, otherwise you'll have strange results.

mikekryjak: Equation is missing the pressure gradient term.

Also, could we rearrange this a bit for clarity? Calculating an intermediate `Nn_eq` makes it look like we are calculating density, which can be confusing on first read. I would prefer if we calculated `NVn` straight away, even if it's a long equation. Don't forget the mass factor as per the comment below.

src/neutral_mixed.cxx
```diff
@@ -35,26 +35,56 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
...
   auto& options = alloptions[name];
 
-  // Evolving variables e.g name is "h" or "h+"
-  solver->add(Nn, std::string("N") + name);
-  solver->add(Pn, std::string("P") + name);
-
   evolve_momentum = options["evolve_momentum"]
                         .doc("Evolve parallel neutral momentum?")
                         .withDefault<bool>(true);
+  passive_momentum = options["passive_momentum"]
```
mikekryjak: I know that this name is now used also in the PR name, but I think there is an opportunity here to better communicate what this model is doing.

In effect, this model replaces the neutral momentum equation with a pressure-diffusion equation, just like what we use for cross-field transport: it's a balance of the neutral pressure gradient and momentum channels from the reactions (CX, IZ and REC). 

If we called this option `parallel_pressure_diffusion`, then it would be clear that the pressure-diffusion model is used to replace the momentum equation. What do you think? 

When I was reading `passive_momentum` I was always assuming this means `Vn = Vi`, because that's kind of what it sounds like to me.

Vandoo:
Update the comment according to the new pattern:

### Purpose

Add a reduced "parallel pressure-diffusion" neutral model to `neutral_mixed`, so a
neutral species can be run by evolving the density only, while the parallel neutral
flow is obtained from a quasi-static parallel momentum balance instead of a dynamical
momentum equation. This avoids resolving fast neutral parallel dynamics, giving a (possible)
cheaper neutral model for edge plasma simulations where the neutral parallel
motion is close to a pressure-diffusion balance (Bufferand et al. 2024, eqs 3-5).

replace the earlier `neutral_passive_momentum` PR (closed); re-implemented on current
master as branch `passive_momentum_v2`.

### Change Summary

- New option `[<neutral>] parallel_pressure_diffusion` (default `false`) in
  `neutral_mixed`. When `true`:
  - only `Nn` is evolved; `evolve_momentum` and `evolve_pressure` default to `false`
    (setting `parallel_pressure_diffusion=true` together with `evolve_momentum=true`
    is rejected with a clear error);
  - `NVn` is set algebraically each RHS from the quasi-static balance
    `NVn = AA * ( Nn_eq * U - (Dnn/Tn) * Grad_par(Pn) )`, i.e. an ion-drag
    convective part plus a parallel pressure-gradient diffusive part, with
    `Nn_eq = (Nn*nu_cx + Ne*nu_rec) / (nu_cx + nu_iz + Rnn)`;
  - the neutral temperature is taken from another species via a new
    `temperature_from` option (e.g. `temperature_from = d+`), which is required in this
    mode; `Pn = Nn * Tn`.
- Added a standalone `evolve_pressure` option to `neutral_mixed` (previously `Pn` was
  always evolved), so the density-only mode is possible.
- Option renamed from `passive_momentum` to `parallel_pressure_diffusion`, and the
  `evolve_momentum` default and a `temperature_from` guard added, per review comments.
- Assumes quasineutrality (Ni = Ne) in `Nn_eq` for now.

### Validation

- Builds and runs on ARCHER2 (PrgEnv-gnu / GCC) and Pitagora against current master.
- Exercised on the MAST-U connected-double-null `nn_cd` case (132x28) in density-only
  mode (`evolve_pressure=false`, `evolve_momentum=false`, `temperature_from=d+`):
  multi-day restarted runs are stable, fields remain finite, and `Td` tracks `Ti` as
  intended.
- Note: in early-phase testing the neutral behaviour differs from the full neutral
  model - divertor electron density (Ne_div) is somewhat lower, Te and Ti are similar,
  neutral density (Nd) is lower and neutral temperature (Td) is higher. This is expected
  given the reduced parallel closure, but is flagged for reviewers as a difference to be
  aware of when substituting this mode for the full model.

### AI Assistance

- An LLM assistant was used to port the original `neutral_passive_momentum` branch onto
  the more recent master (as `passive_momentum_v3`), and to help debug the errors found
  during review (e.g. those Mike identified). All resulting changes were reviewed and
  tested by the author before inclusion.

### Documentation

- Option documentation is provided inline via the `.doc(...)` strings for
  `parallel_pressure_diffusion`, `evolve_pressure`, and `temperature_from`.

### Review Notes

- `Nn_eq` currently assumes `Ni = Ne`.
- `temperature_from` is mandatory in this mode; the code errors clearly if it is unset.

PR #596, opened 2026-06-29
Add metadata to component objects

cmacmackin:
## Purpose

It could be difficult to produce meaningful error messages, because component names are not consistently stored and their type information is not available at runtime. This PR provide a standard interface for this metadata.

## Change Summary

The name of the component object is now a mandatory parameter for the `Component` base class. This gets stored in a private variable and can be accessed using the public `objectName()` method.

The component type names were a bit more complicated, as it is useful to have these available both statically and dynamically. Components are now expected to provide a `static constexpr auto type` component which can be converted to a string. This is used by the `RegisterComponent` helper type to set the type-name associated with the component class (as opposed to passing it to the constructor manually). Additionally, all `Component` classes must implement the virtual function `typeName()`.

For most sub-classes, this is done using the [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern). Rather than inheriting directly from `Component` they inherit from `template<typname T> NamedComponent`. The secondary `NamedComponent` class was required so that method implementations for the `Component` base class can still be put in a separate file from the header. Note that the reaction classes create many components from the same (possibly templated) class and therefore don't use `NamedComponent`. Instead they use custom logic to set the static constant and define the virtual method.

A formatter has been defined for `Component` objects, which makes use of the `objectName()` and `typeName()` methods. By default, this prints names with the format `objname (typename)`, or just `typename` if these two names are the same.

## Validation

New unit tests were written for the `objectName()` method and the formatter, which confirms that these work as intended. 

Additionally, a unit test was added which instantiates all component type registered with the factory and confirms that `typeName()` returns the same name as used to create them with the factory. Note that this is the first time there has been any test that many of these components can be constructed at all.

## AI Assistance

No AI tools were used.

## Documentation

The developer documentation has been updated to reflect the new requirements when writing a component class. I will leave it up to reviewers to advise whether they think any additional details would be beneficial.

## Review Notes

This work was originally done on top of #428, but I've spun it off so it can be merged earlier.


cmacmackin:
Irritatingly, the coverage test is failing because `prek` and `clang-format` went and totally rewrote a few files. The changes were only to formatting, but they are showing up as untested.

cmacmackin:
@ZedThree I know you're busy but I thought you might be a good person to review this, as it involves some messing around with templates and meta-programming.

mikekryjak:
Changes and docs look straightforward to me but I'm afraid I lack the C++ chops to give this a proper review.

dschwoerer:
I fail to see why we need CRTP here. Why do you want typename to be a constexpr?
I mean, sure, if it is no bother, that is nice, but I feel like this could also be a normal, virtual member function.
No templating needed then.

I think CRTP is great, as it allows to compiler to inline code, but it has a cost, that everything is a template, and that make compiling more expensive. Surely typeName should never be used in a hot loop. Could you explain what is the advantage? A quick grep of the code base did not make it obvious that typeName is used somewhere in a constexpr.

cmacmackin:
The reason I want the type name accessible at compile-time is so that we can guarantee it is the same as the value used in the `RegisterComponent` constructor. I don't want to run into a situation where someone could change the type name used in `typeName()` but not the one used in the `BOUT.inp` files to instantiate the component object. You don't necessarily need a `constexpr` to achieve that--it could be done via a `static` method instead--but crucially you do need both a static and dynamic way to access the type-name.

cmacmackin:
I've added some tests for `FixedFractionRadiation` and `SheathBoundarySimple`, to get the patch coverage up. The tests for the latter are limited. I just copied the tests from `SheathBoundary` and kept the ones that still passed for `SheathBoundarySimple`. Someone that understands the physics for `SheathBoundary` and `SheathBoundarySimple` should add more tests which can actually account for the differences between them.

bendudson (approved):
Thanks, @cmacmackin !
It may be easiest to just add an option to enable/disable sorting in #428 and merge that.

cmacmackin:
Given both this commit and #428 contain quite substantial changes, I feel like it would be best to keep them separate for ease of reviewing.

include/solkit_hydrogen_charge_exchange.hxx
```diff
@@ -53,21 +55,25 @@ struct SOLKITHydrogenChargeExchangeIsotope : public SOLKITHydrogenChargeExchange
     substitutePermissions("readvals", {"AA", "density"});
   }
 
+  //FIXME: Need to figure out how to consturct type name at compile-time
+  static constexpr char type[] = {'s', 'o',     'l', 'k',     'i', 't', ' ',     Isotope,
```
ZedThree, line 58: There's a new `join` template in this PR that says it does compile-time concatenation, can it be used here?

ZedThree (approved):
Looks great, thanks @cmacmackin! Just one comment on a `FIXME`

include/solkit_hydrogen_charge_exchange.hxx
cmacmackin: I think in this case it's simpler just to define a character array than to mess around with the` join` template. The latter would require that I define a bunch of string_view objects, including for `Isotope`, which doesn't seem worth the hassle. I'm just going to delete the FIXME comment.

PR #601, opened 2026-07-01
Update docker builds to work for multi-arch, with smart reuse of built packages

tbody-cfs:
**Multi-arch Docker images + faster, reproducible builds.**

Makes `hermes-3`, `hermes-3-builder`, `hermes-3-jupyter` multi-arch (`amd64`+`arm64`) so Apple-silicon/arm64 hosts get native images instead of failing under emulation.

### Build strategy
- Per-arch native runners (no QEMU for Spack), push-by-digest, merge into one manifest. Jupyter built multi-arch via QEMU (apt/pip only).
- Two-tier Spack binary cache: public mirror + self-hosted OCI cache (`ghcr.io/boutproject/hermes-3-spack-cache`), keyed per-package hash. Cuts builder install ~30 min → ~90 s.
- `spack install` no longer `--fail-fast` in the cache branch; partial results are pushed regardless of exit, then the exit code is re-propagated so CI still fails.

### Portability pin
- Hard per-arch `packages:all:require:target=` (`aarch64` / `x86_64_v3`) injected in the builder Dockerfile. `granularity: generic` is only a *preference* — with cache reuse the concretizer pulled `armv9.0a`/AVX-512 binaries that SIGILL'd on older CPUs / Docker VMs.

### CI
- New `test_docker_build.yml`: on `docker/**` PRs, builds the builder→final (+jupyter) chain on both arches under the `experimental_docker_build` tag, leaving `latest`/`edge` untouched. Skips cleanly on fork PRs.
- Added release trigger to `build_docker_image.yml`; explicit `permissions` + concurrency guards on all publishing workflows; OCI cache wired into the production builder.

### Other
- Base images → Ubuntu 24.04 (noble), maintained `quay.io/jupyter/scipy-notebook`, all pinned by digest. Builder drops the apt layer (base already ships the GCC toolchain; cmake built by Spack; gcc registered via `spack external find`).
- `CMakeLists.txt`: examples / json_database / tests copies gated behind new `HERMES_COPY_*` options (default `ON`, so no change for normal builds); the docker `hermes_config.cmake` disables the examples + tests copies to slim the image.
- `run.sh` compose helper; `HERMES_TAG` / `JUPYTER_TAG` in compose + `.env`; dockerignore renamed to the effective `hermes-3.dockerfile.dockerignore`; dropped the removed `enable_c.patch`; README rewritten to match the actual two-Dockerfile setup.

Built / AI-assisted with Claude Code; concretization + end-to-end `work/test` solve verified on rebuilt CI images (arm64 native + amd64 emulated). `x86_64_v3` runtime re-validation is pending a fresh build (end-to-end solve was verified at the earlier `v2` pin).


tbody-cfs:
The multi-arch image can be pulled from https://github.com/boutproject/hermes-3/pkgs/container/hermes-3. Look for the `experiment_docker_build tag`.
On my Apple Silicon Mac, I can pull and run the arm64 image.

Currently this is failing to run hermes, due to illegal instructions. Might be arm_v9 build vs arm_v8 on my laptop. I'll investigate.

tbody-cfs:
Fixed the illegal instructions issue!
Can now run the `examples/tokamak-1D/1D-threshold/BOUT.inp` case in 58s on my laptop using the arm image, or in 1m36s using emulation with the amd image.

dschwoerer:
If they images are to be used in production, we likely want x86_64-v3, not x86_64-v2.

Optimisation efforts would otherwise be in vain, if we do not even try to utilize the hardware.

tbody-cfs:
> If they images are to be used in production, we likely want x86_64-v3, not x86_64-v2.
> 
> Optimisation efforts would otherwise be in vain, if we do not even try to utilize the hardware.

Hi @dschwoerer,
Sure thing — I can bump the target to v3. However, I won't be able to test on my arm laptop then. Could someone with an x86 chip pull the `experimental_docker_build` tag and test it? You should be able to make a directory `docker/work/test` and put a `BOUT.inp` in it. Then run `sh setup.sh` and
```HERMES_TAG=experimental_docker_build docker compose run --rm hermes work/test```
I've confirmed this works for the arm image, but need the x86 tested

If there's significant interest in targetting different archs, we could also release more specific separate tags targetting optimized chip archs. For instance `latest` could point to a portable `arm_v8/x86_64_v3`, but if a user wanted they could also pull a `arm_v9` or `x86_64_v4` tag. Would be a follow-up PR and non-zero effort, but I can look into this if there's sufficient enthusiasm to actually use the images.

As a side note: https://github.com/boutproject/hermes-3/pkgs/container/hermes-3-spack-cache is now available for any other Spack builds in `boutproject`. Feel free to reuse these!

**Comment by tbody-cfs** [comment]
Looks like the spack cache is working its magic! In a recent run on master, `build-builder-image` took 1h 22m (see https://github.com/boutproject/hermes-3/actions/runs/28496789415). In the most recent workflow run for this PR, `build-builder-image` took 8m (see https://github.com/boutproject/hermes-3/actions/runs/28593806817/job/84783888642)!

Let me know if you want to use this for the BOUT image as well

**Comment by tbody-cfs** [comment]
```
## Claude review of code-correctness and security

Reviewed all 17 changed files adversarially against the merge-base. **No security issues, no regressions, no broken code.** One stale doc line fixed (see below).

### Security — clean
- `test_docker_build.yml` uses plain `pull_request` (not `pull_request_target`): fork PRs get a read-only token and never run privileged. Fork builds are gated off via `head.repo.full_name == github.repository`; the merge/publish jobs skip on forks through their `needs:` chain.
- No script injection — no free-form `${{ github.event.* }}` interpolated into `run:` blocks. `github.actor` / `SPACK_OCI_USER` are usernames, not secrets.
- `ghcr_token` is a BuildKit `type=secret` mount, read into an env var only within a single `RUN` (never an `ENV`/build-arg) → not baked into any image layer. Non-empty guarded with `-s`.
- `run.sh`: service names validated against an allowlist before `docker compose run … "$@"`; all vars quoted; no `eval` / `curl|bash`. Compose bind-mounts only `./work`, no privileged/socket mounts.

### Correctness — verified sound
- Exit-code re-propagation `{ spack install; rc=$?; buildcache push || true; exit $rc; }` is correct: partial results still pushed, CI still fails on error.
- Manifest merge (digests → `imagetools create`), job dependency graph, permissions, and concurrency groups all check out. The README workflow-trigger table matches the actual `on:` blocks.
- `.git` correctly retained by the dockerignore (needed for in-container submodule init).

### Fixed
- `docker/README.md` said jupyter builds `FROM jupyter/scipy-notebook`, but the Dockerfile now uses `quay.io/jupyter/scipy-notebook`. Corrected.

### Verify before merge (could not confirm locally)
1. **`spack config add "packages:all:require:target=…"`** — that this appends to the existing list-form `require: ['%gcc']` (→ `['%gcc', 'target=x86_64_v3']`) rather than clobbering it into an invalid dict. Spack's documented behaviour appends to a list; confirm via `spack -e . config get packages` in a build. Note the PR flags `x86_64_v3` runtime re-validation as pending — the end-to-end solve was verified at `v2`, not yet `v3`.
2. **The renamed dockerignore is now active for the first time** (the old name never matched the Dockerfile, so it was inert). All patterns — examples, googletest, docs — now apply. Consistent with `HERMES_TESTS=OFF` / `HERMES_COPY_*=OFF`, and the CI `work/test` solve passed, but correctness rests on that build staying green.
```

**Comment by bendudson** [comment]
I'm probably not qualified to review this properly, but I think it looks great. Thanks very much @tbody-cfs !

**Comment by mikekryjak** [comment]
Sorry, I've not yet played with Docker so I'm not the best person to review. @mrhardman who has implemented docker builds in our VANTAGE fork and I know @dschwoerer has also done some Docker work before.

File: .github/workflows/test_docker_build.yml [hunk 1]
```diff
@@ -0,0 +1,373 @@
...
+#     emulation) and published as 'experimental_docker_build' too. It is
+#     independent of the builder->final chain.
+#
+# NOTE: publishing requires write access to the registry, so this only works for
+# branches in this repository, not pull requests opened from forks.
+on:
+  pull_request:
+    paths:
+      - 'docker/**'
+      - '.github/workflows/test_docker_build.yml'
```
dschwoerer, line 28:
```suggestion
      - '.github/workflows/test_docker_build.yml'
      - '.github/workflows/build*image.yml'
```

.github/workflows/test_docker_build.yml
```diff
@@ -0,0 +1,373 @@
...
+      (
+        (needs.changes.outputs.builder == 'true' && needs.merge-builder.result == 'success') ||
+        (needs.changes.outputs.builder != 'true' && needs.changes.outputs.final == 'true')
+      )
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-latest
```
dschwoerer: Maybe use `ubuntu-24.04` - to ensure that both arches stay aligned?
It is a pain there is no `ubuntu-latest-arm`.

docker/image_ingredients/docker_image_commands.sh
```diff
@@ -79,7 +79,7 @@ build_boutpp () {
     check_exit_code "BOUT++ configuration failed"
 
     notice "Finished configuring BOUT++. Starting build"
-    cmake --build "${build_dir}" --parallel
+    cmake --build "${build_dir}" --parallel 4
```
dschwoerer: It is probably good to limit the number of builders, but I think we only have two cores. For BOUT++ and Hermes we are typically CPU limited, due to templates, so I would assume 2 or 3 should work better.

docker/hermes-3-builder.dockerfile
```diff
@@ -1,30 +1,99 @@
...
-RUN spack compiler find && spack external find cmake
+# No OS packages are installed here: the base image already ships the full GCC
+# toolchain (gcc/g++/gfortran/make, libc6-dev) plus git, and cmake is built by
+# Spack as part of the environment (listed in spack.yaml).
+#
+# Register gcc as an external package. It is already registered by the base
+# image in /root/.spack/packages.yaml; re-running is harmless and acts as a
+# safety net. In Spack 1.x compilers are graph nodes and the concretizer only
+# accepts compilers found as external *packages* via `spack external find` (the
+# legacy `spack compiler find`, which writes compilers.yaml, is ignored).
```
dschwoerer: please shorten

docker/hermes-3.dockerfile
```diff
@@ -54,7 +58,7 @@ RUN . /opt/spack-environment/activate.sh \
           -S ${BOUTPP_SRC_DIR} \
           -C ${BOUTPP_CONFIG} \
           -Wno-dev \
-&& cmake --build ${BOUTPP_BUILD_DIR} --parallel 2
+&& cmake --build ${BOUTPP_BUILD_DIR} --parallel 4
```
dschwoerer: I would expect this makes things slower in CI. It might be nice to have this as an option, and default to 4 (?) and override it in CI to 2.

CMakeLists.txt
```diff
@@ -254,23 +254,30 @@ else()
...
+      $<TARGET_FILE_DIR:${PROJECT_NAME}>/json_database)
+endif()
+
+if(HERMES_COPY_TESTS_TO_BUILD)
+  add_custom_command(
+    TARGET setup-test
+    PRE_BUILD
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_SOURCE_DIR}/tests
+            $<TARGET_FILE_DIR:${PROJECT_NAME}>/tests)
+endif()
```
dschwoerer: This seems wrong. They are extra targets, that should always be generated.

docker/run.sh
```diff
@@ -0,0 +1,188 @@
...
+    *) return 1 ;;
+  esac
+}
+
+run_docker_help() {
+  notice "Usage: run_docker <service> [arguments]"
+  echo
+  echo "Available services (from docker-compose.yaml):"
+  echo "  shell            Interactive terminal shell in the image"
+  echo "  sudo             Interactive shell with root access"
```
dschwoerer:
```suggestion
  echo "  shell            Interactive shell in the image"
  echo "  sudo             Interactive shell with root access"
```

docker/README.md
```diff
@@ -81,65 +142,117 @@ Here's a breakdown of the available commands:
...
+
+```sh
+case "$(uname -m)" in
+  aarch64) HERMES_TARGET=aarch64 ;;
+  x86_64)  HERMES_TARGET=x86_64_v3 ;;
+esac
+spack -e . config add "packages:all:require:target=${HERMES_TARGET}"
+```
+
+`aarch64` and `x86_64_v3` are portable baselines (`x86_64_v3` ≈ any x86-64 CPU since ~2013 — Intel Haswell / AMD Excavator and later — adding AVX2, BMI1/2 and FMA). Requiring a *generic-family* target this way does **not** break concretization of the externally-found gcc — a newer compiler can always target an older ISA — despite the compiler being a graph node in Spack 1.x. (A more specific microarch, e.g. `x86_64_v4`, could conflict with the external gcc's detected target; the generic family does not.)
```
dschwoerer: This is unfortunately not true. Intel thought it would be great to make more money, so they shipped non -v3 CPUs until very recently.

However, let's hope users that try to run hermes-3 have a better CPU ...

dschwoerer (commented):
Personally I find the comments way to long.
You did not mention that you used AI, but it certainly feels AI generated. Could you maybe review the comments, and shorten them?

docker/hermes-3-builder.dockerfile
tbody-cfs: I've shortened the comments (via a combo of hand-editing and telling Claude to be more succinct). Particularly for spack stuff, I prefer leaving fairly detailed comments since the documentation is sometimes non-obvious (at least, it was when I dug into it in the past).

.github/workflows/test_docker_build.yml
tbody-cfs: Done. Now using `ubuntu-24.04` (amd64) alongside `ubuntu-24.04-arm`, across all three build workflows.

docker/image_ingredients/docker_image_commands.sh
tbody-cfs: Made a new `HERMES_BUILD_JOBS` environment variable, which can be set either in `.env` or via the CLI. Defaults to 4, but overridden to 2 for the CI/CD pipeline.

docker/hermes-3.dockerfile
tbody-cfs: Using 2 jobs in CI via the `HERMES_BUILD_JOBS` argument

docker/README.md
tbody-cfs: Sad :( I removed that text. Also saw your https://github.com/boutproject/hermes-3/pull/603 PR — we could make a x86_64_v2 tag for compatibility, if desired.

CMakeLists.txt
tbody-cfs: This one is a tricky one. I was getting a CI failure (see https://github.com/boutproject/hermes-3/actions/runs/28533905473/job/84598731624).

tbody-cfs: If you want to reproduce this yourself, you can try
```bash
git checkout docker_fixes
git checkout master -- CMakeLists.txt
docker pull ghcr.io/boutproject/hermes-3-builder:experimental_docker_build
# This will fail
docker build \
  -f docker/hermes-3.dockerfile \
  --build-arg BUILDER_TAG=experimental_docker_build \
  -t hermes-3:cmake-test .
```
The error will be something like
```
-- Build files have been written to: /hermes_project/build/hermes-3-build
Error copying directory from "/hermes_project/src/hermes-3/examples" to "/hermes_project/build/hermes-3-build/examples".
gmake[2]: *** [CMakeFiles/setup-examples.dir/build.make:73: setup-examples] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:158: CMakeFiles/setup-examples.dir/all] Error 2
...
ERROR: failed to build: ... exit code: 2
```
Happy for any other ideas how to fix/why this is happening

tbody-cfs: So dug into this a bit more. It seems like newer versions of CMake fail if you try to copy an empty source directory. This should fail for anything after v3.15 (released 2019), so I'm surprised that it hasn't come up as an issue before. However, if I change the guard around the copy to a simple `IF(EXISTS(` (see https://github.com/boutproject/hermes-3/pull/601/changes/2320fb3e4bdd29ffbf277821a7dd05e70fd6512b) then the pipeline still runs through. Therefore, I'm pretty sure that these clauses should only be run if those flags are true (but happy to keep the `IF(EXISTS` which does the same thing).

See docs at https://cmake.org/cmake/help/latest/manual/cmake.1.html
```
copy_directory -t <destination> <dir>...
Copy content of <dir>... directories to a <destination> directory. If the <destination> directory does not exist it will be created. If -t is not specified, the last argument is assumed to be the <destination>. copy_directory does follow symlinks.

Added in version 3.5: Support for multiple input directories.

Added in version 3.15: The command now fails when the source directory does not exist. Previously it succeeded by creating an empty destination directory.
```

docker/README.md
```diff
@@ -81,65 +158,149 @@ Here's a breakdown of the available commands:
...
+
+```sh
+case "$(uname -m)" in
+  aarch64) HERMES_TARGET=aarch64 ;;
+  x86_64)  HERMES_TARGET=x86_64_v3 ;;
+esac
+spack -e . config add "packages:all:require:target=${HERMES_TARGET}"
+```
+
+`aarch64` and `x86_64_v3` are portable baselines (`x86_64_v3` adds AVX2, BMI1/2 and FMA on top of the base x86-64 ISA). Requiring a *generic-family* target this way does **not** break concretization of the externally-found gcc — a newer compiler can always target an older ISA — despite the compiler being a graph node in Spack 1.x.
```
dschwoerer:
```suggestion
`aarch64` and `x86_64_v3` are portable baselines (`x86_64_v3` adds AVX2, BMI1/2 and FMA on top of x86-64_v2 ISA). Requiring a *generic-family* target this way does **not** break concretization of the externally-found gcc — a newer compiler can always target an older ISA — despite the compiler being a graph node in Spack 1.x.
```
I would guess AVX2, BMI1/2 and FMA are all not in v2.

dschwoerer (approved):
I think it looks good. I have not read the documentation, as it is still to long, in my opinion. Others might disagree, so I am fine with it.

dschwoerer:
One thing, I think it would be good to have the examples in the container. They are not large, and if you want to start playing with hermes-3, that makes it much easier if you can start working with the examples.

tbody-cfs:
I've added the examples in [6078f18](https://github.com/boutproject/hermes-3/pull/601/commits/6078f18d0beb4c8907ccd098742511f626e26fa8)

PR #604, opened 2026-07-03
Optimise runtime with collision, ADAS, div_ops, and state-management improvements

Quarrinexus:
### Purpose

To reduce hermes-3 runtime cost via a set of internal optimisations to the collision-frequency calculation, ADAS reaction-rate evaluation, div_ops metric handling, and per-RHS-evaluation state management — with no change to simulation physics or user interfaces. No new physics capability.

### Change Summary

Two classes of change, verified separately (different correctness bar for each, see Validation):

**1. Floating-point reordering (mathematically identical, not bit-identical):**
- `braginskii_collisions.cxx` / `braginskii_ion_viscosity.cxx`: hoists scalar (per-species) constants out of the collision-frequency loops; replaces `pow(x, 1.5)` with `x * sqrt(x)`.
- `adas_reaction.cxx` / `integrate.hxx`: Introduces cellAverageInto() so that OpenADAS and OpenADASChargeExchange reuse a persistent Field3D workspace instead of constructing a fresh one on every evaluation call. 

**2. Framework changes (no algebraic reordering at all, bit-identical):**
- `div_ops.cxx`: Replaced an expensive modulo operator (%) used for boundary wraparound with a standard if/else conditional branch. Because this logic only handles integer grid indices, the output is identical.
- `permissions.cxx` / `guarded_options.*`: Optimises access-rights lookup and memory tracking. It now remembers the results of bestMatchRights and replaces high-overhead, per-construction tracking maps with a more lightweight session-ID scheme.
- `div_ops.cxx`: Caches grid metric calculations within the `Coordinates` object, the dominant performance gain in this PR, avoiding repeated recalculation and shifting of field-aligned components on every flow evaluation.
- `hermes-3.cxx`: the state `Options` tree persists between RHS evaluations instead of being rebuilt every call.
- `neutral_full_velocity.cxx`: merges three `mesh->communicate()` calls into one.
- New `hermes:check_state_values` option (default true) to skip the per-`set()` finiteness sweep when unneeded.

### Validation

Verified on Imperial College London's CX3 HPC across all 10 integrated tests using `scripts/compare_builds.py` (new file, included in this pull request), which runs both binaries simultaneously and reports per-variable field differences as well as wall-clock timing. 
- Class 1 changes were checked with a ULP tolerance (a few ULPs / ~1e-15 relative expected from valid reordering). 
- Class 2 changes were checked bit-for-bit (every variable, every timestep, any nonzero difference would be a bug).

Master vs. this branch (commit 7f0ae4d), single PBS job across all 10 tests, 5 timed runs per binary per test (note compare_builds.py defaults to 3):

| Test | OLD (mean) | NEW (mean) | Speedup |
|---|---|---|---|
| `neutral_mixed` | 9.73s | 7.47s | **+30.2%** |
| `1D-recycling` | 34.52s | 29.73s¹ | **+16.1%¹** |
| `drift-wave` | 11.90s | 10.80s | +10.1% |
| `1D-recycling-dthe` | 28.54s | 27.69s | +3.1% |
| `2D-recycling` | 3.43s | 3.34s | +2.7% |
| `alfven-wave` | 10.00s | 9.92s | +0.8% (noise) |
| `collfreq-braginskii-afn` | 2.46s | 2.46s | 0.0% (noise) |
| `vorticity` | 2.48s | 2.48s | 0.0% (noise) |
| `collfreq-multispecies` | 2.46s | 2.46s | -0.1% (noise) |
| `2D-production` | 3.41s | 3.46s | -1.4% (noise) |

¹ One of the 5 timed runs for the new binary spiked to 121s against the other 4's tight 29.6-29.9s (scheduler/network hiccup, not reproducible). NEW (mean) above and the speedup figure both exclude that outlier.

Caching the grid metrics in div_ops provides the largest performance boost. Diffusion-heavy simulations (like `neutral_mixed`) show a ~25-30% speedup in production (CHECK=0): the exact figure has varied between separate job submissions on this shared HPC queue (11.8% in an earlier 3-run job vs. 30.2% here), but this test is consistently the largest and most reproducible win across every comparison in this investigation. The speedup will be even greater in development builds (CHECK≥1) because this cache also eliminates a repetitive string scan and a full-field validation sweep that are normally compiled out of production runs.

No correctness bug was found in the actual code changes. `compare_builds.py` did surface differences on some tests, but each traces to one of two known, explained causes rather than being hidden; see Review Notes for the full details on these and other caveats.

Run it yourself:
```
python3 scripts/compare_builds.py --old /path/to/master/hermes-3 --new /path/to/branch/hermes-3 --tests-dir tests/integrated --mpirun "mpirun -np"
```

### AI Assistance

* **AI Contributions (Claude - Sonnet/Fable):** 
    * Identified the initial performance bottlenecks and proposed optimisation strategies.
    * Assisted with the implementation and integration of the code changes.
    * Wrote, tested, and iterated the `scripts/compare_builds.py` verification tool.
    * Provided troubleshooting support for compilation and build-system errors encountered along the way.
* **Human Verification & Oversight:** 
    * Directed the overarching investigation and defined the testing methodology (establishing the boundaries for bit-exact vs. mathematical rounding tolerances).
    * Audited and filtered out incorrect or counter-productive performance suggestions made by the AI.
    * Reviewed the High-Performance Computing (HPC) cluster logs, identified anomalous test runs, and managed necessary reruns.

### Documentation

* `scripts/compare_builds.py` documents itself via its module docstring and `--help` (meaning users can view usage instructions directly in the terminal).
* The README / existing manual (`docs/sphinx/`) was not changed for the following reasons:
    * **The new setting explains itself automatically:** `hermes:check_state_values` is self-documenting. It hooks into the codebase's existing `.doc()` feature to automatically display its own description text to users when requested.
    * **No existing documentation is out of date:** A search was done for references to every modified source file in this PR. The manual only mentions these files to describe high-level component names or physics theories, none of which were altered.
    * **The changes are all behind-the-scenes:** Every single modification is an internal implementation detail. Because the underlying physics, user configurations, and public interfaces remain completely identical, there is no new behaviour for a user manual to describe.

### Review Notes

- **`1D-recycling`/`1D-recycling-dthe`** have a density-feedback controller that chaotically amplifies any bit-level rounding change late in the run; a valid FP reordering can land the density trajectory on the opposite side of the controller's threshold, after which the two runs integrate different (both valid) trajectories. `compare_builds.py` compares an earlier timestep for these two tests specifically, and still reports the residual differences.
- **Diagnostic x-boundary guard cells** (`SNVd+`, `ddt(Nd)`, etc., on `2D-production`/`2D-recycling`) are never written by Hermes-3, so their output is whatever stale allocation content happens to be there; this differs between builds whenever the allocation sequence changes (e.g. the div_ops metric cache), unrelated to correctness. `compare_builds.py` detects and classifies this rather than reporting a false failure. Might be worth a follow-up to zero those cells on write so output is byte-reproducible in general.
- **Not tested against the current upstream version.** This branch's fork point predates ~20 commits on `master`. The diff still applies and merges without conflict, but the correctness/performance verification above is relative to the fork point, not current `master`. (tested in later comment)
- **`neutral_full_velocity`'s merged `communicate()`** has no integrated-test coverage; its correctness argument is structural only (guard-cell exchanges of independent fields commute), not verified by a test in this PR.
- **CHECK≥1 gains are unmeasured.** The CX3 build used CHECK=0, which compiles out the permissions/GuardedOptions/finiteness-sweep code paths these changes optimise, so their benefit is invisible in the numbers above and should be larger on a development (CHECK≥1) build. (tested in later comment)
- **`hermes:check_state_values=false` is untested.** `compare_builds.py` has a dedicated `--verify-check-state-values` flag for exactly this (runs the NEW binary once more with the option disabled and requires bitwise-identical output), but it was never invoked in this investigation, it's only meaningful on a CHECK≥1 build, and every build here used CHECK=0, where the finiteness sweep the option skips is already compiled out. So beyond ensuring the change's existence doesn't cause any issues, the option's actual claim (disabling it is safe and saves time) has no test coverage yet; verifying it would need a CHECK≥1 build. (tested in later comment)


include/integrate.hxx
```diff
@@ -81,16 +81,43 @@ auto cellAverage(Function func, const RegionType &region) {
...
     return result;
   };
 }
 
+/// Like cellAverage but writes into a caller-supplied Field3D, avoiding a heap
+/// allocation on every call. On the first call result is initialised from the
+/// first input field; subsequent calls reuse the existing allocation.
+template <typename CellEdges = hermes::Limiter, typename Function, typename RegionType,
+          typename... Fields>
+void cellAverageInto(Field3D& result, Function func, const RegionType& region,
```
bendudson, line 61: Should `cellAverage` call `cellAverageInto` to avoid duplicating code?

Note that `Field3D` data is not heap-allocated when new Field3D objects are created: The backing `Array` class uses a per-thread arena allocator so the cost is a map lookup (key is array size) and popping a pointer from a list.

include/integrate.hxx
```diff
@@ -81,16 +81,43 @@ auto cellAverage(Function func, const RegionType &region) {
...
+                     const Fields&... args) {
+  if (!result.isAllocated()) {
+    result = emptyFrom(firstArg(args...));
+    result.allocate();
```
bendudson: I think `result.allocate()` should always be called, so moved after this `if`. Fields use copy-on-write so if `result` shares data with another field then modifying it could have non-local effects. `allocate()` ensures that the backing data is unique and is a cheap call if it is already unique.

include/integrate.hxx
Quarrinexus: Good catch, I've implemented this in commit 4cda33e8.

Thanks for the explanation of the array store, I have changed the description to correct my error.

include/integrate.hxx
Quarrinexus: I've implemented this in commit 4cda33e, thanks for the tip.

Quarrinexus:
I've now run scripts/compare_builds.py between current upstream master and the rebased branch, both built with -DCHECK=1, 5 timed runs per test (1 warmup), across all 10 integrated tests including 2D-production/2D-recycling (nproc=10, MPI). Since this branch is now rebased onto current master, this also removes any potential doubt from the previous comparison having been run against a base several commits behind upstream. This confirms that setting hermes:check_state_values=false produces bitwise-identical output to the default (true) across every test, including under MPI. The results are below.

| Test | OLD (mean) | NEW (mean) | Speedup |
|------|-----------|-----------|---------|
| `collfreq-braginskii-afn` | 3.09s | 2.17s | **+42.2%** |
| `neutral_mixed` | 9.27s | 7.11s | **+30.5%** |
| `1D-recycling` | 39.16s | 33.58s | **+16.6%** |
| `alfven-wave` | 10.93s | 9.77s | +11.9% |
| `drift-wave` | 11.71s | 11.05s | +6.0% |
| `1D-recycling-dthe` | 33.01s | 31.76s | +3.9% |
| `vorticity` | 2.27s | 2.25s | +1.1%  (noise) |
| `collfreq-multispecies` | 2.15s | 2.16s | -0.5% (noise) |
| `2D-production` | 3.68s | 3.74s | -1.6% (noise) |
| `2D-recycling` | 3.53s | 3.81s | -7.5% (noise) |

**CHECK≥1 gains**: measurable and generally larger than the CHECK=0 numbers in the PR description, since `check_state_values` removes CHECK-gated overhead that CHECK=0 never paid. The two nproc=10 MPI tests initially showed small slowdowns, consistent with MPI scheduling noise on a shared node rather than a real regression (comparable in size to `collfreq-multispecies`'s -0.5% at nproc=1). This was confirmed by re-doing the 2D-recycling test 20 times: the slowdown reversed to an average speedup of 1.8%.

**`hermes:check_state_values=false` bitwise safety**: confirmed on **all 10 tests**, including both MPI tests, every run produced bit-identical output with the finiteness sweep disabled.

**3 ULP-tolerance failures** (1D-recycling, 1D-recycling-dthe, drift-wave): pre-existing, explained in the intial PR-description.

**2D production/2D recycling ULP-tolerance failures:** Only one variable, `efd+_tot_ylow`, exceeded the 10-ULP tolerance, reaching 24 ULPs. All other variables were within the tolerance, or matched the previously known x-boundary guard-cell differences, which are caused by unused memory rather than the simulation itself. This is consistent with expected small floating-point differences from MPI framework changes and does not indicate a correctness issue.

src/neutral_full_velocity.cxx
```diff
@@ -483,13 +483,10 @@ void NeutralFullVelocity::finally(const Options& state) {
...
-    mesh->communicate(Dnn);
+    // Communicate together to reduce the number of MPI exchanges
+    mesh->communicate(Dnn, kappa_n, eta_n);
     Dnn.applyBoundary();
-
-    mesh->communicate(kappa_n);
     kappa_n.applyBoundary("neumann");
-
-    mesh->communicate(eta_n);
     eta_n.applyBoundary("neumann");
```
dschwoerer: I think the comment is not needed.
Also, we need to first apply BCs, before we communicate, for FCI.
```suggestion
    Dnn.applyBoundary();
    kappa_n.applyBoundary("neumann");
    eta_n.applyBoundary("neumann");
    mesh->communicate(Dnn, kappa_n, eta_n);
```

Quarrinexus: I have applied this fix in the latest commit

PR #603, opened 2026-07-03
Add multiple microarch levels

dschwoerer:
### Validation

CI is running to test it

PR #608, opened 2026-07-07
Add compatibility images

tbody-cfs:
### Purpose
The multi-arch Docker pipeline (#601) publishes images built for x86_64_v3 (AVX2/FMA) on amd64, which fail with Illegal instruction on older x86 CPUs. This adds an older-arch-compatible image variant so users on pre-Haswell hardware can run Hermes-3 in Docker. It also closes a gap in #601: the microarch was only pinned for the Spack dependencies, not the BOUT++/Hermes-3 code itself.

### Change Summary
- Parameterised the microarch. HERMES_TARGET is now a build arg in hermes-3-builder.dockerfile (empty → the previous uname -m default), replacing the hardcoded case.
- Compat image variant. build_builder_image.yml and build_docker_image.yml gained a matrix variant for x86_64_v2, published as suffixed tags (:latest-x86_64_v2 and the edge equivalent for the builder) via metadata-action flavor: suffix. The default multi-arch tags are unchanged.
- Application code now respects the target. hermes-3.dockerfile maps HERMES_TARGET → a GCC -march string, bakes it into /etc/hermes-march, and passes it to the BOUT++/Hermes-3 CMake builds — previously these compiled at GCC's generic baseline regardless of the image's target.
- In-container rebuilds stay consistent. docker_image_commands.sh reads /etc/hermes-march so build_hermes/build_boutpp match the image; standard CMAKE_CXX_FLAGS/CMAKE_C_FLAGS (plumbed through .env → docker-compose.yaml) are appended, letting users add flags or retarget (a later -march wins).

### Validation
- YAML-linted the workflows and docker-compose.yaml; bash -n on setup.sh and docker_image_commands.sh.
- Verified the HERMES_TARGET → -march mapping and the flag-append logic (empty → image default; -march=native and -funroll-loops append correctly) by running the shell snippets standalone.
- Confirmed the config .cmake files set no CMAKE_CXX_FLAGS, so the injected -D flags don't clobber existing settings.
- Exercised end to end in CI via test_docker_build.yml, which builds the default (amd64 v3 + arm64) and x86_64_v2 variants.

### AI Assistance
Developed with Claude Code. It made the dockerfile/workflow/script edits, cross-referenced the parallel PR #603 to pull in the -march-on-final-builds idea (while rejecting its unsafe x86_64_v4 default and armv8.6a ARM floor), diagnosed the CI failure that led to dropping the v1 variant, and validated the shell/YAML locally. All changes were reviewed by me before committing.

### Documentation
Added a "Compiler flags and the per-image -march" section to docker/README.md: a table mapping each tag to its baked -march, and how CMAKE_CXX_FLAGS/CMAKE_C_FLAGS append to it for rebuilds.

### Review Notes
- x86_64 (v1) variant removed due to a failing build. It was originally included alongside v2, but its builder leg fails in CI: v1 is a cold-cache target (nothing prebuilt in the OCI Spack cache), so it must build readline@8.3 from source, and Spack cannot fetch the required readline83-00x patches — they 404 on binaries.spack.io/develop's source-cache and the only fallback (ftpmirror.gnu.org → mirror.latigo.net) is unreachable from the runners. The v2 variant is unaffected because it pulls readline prebuilt from the cache. v1 can be re-added once the readline@8.3 patches are mirrored (or by pinning readline to a mirrored version, e.g. readline@8.2, in spack.yaml).
- Bootstrap ordering: the compat final image builds FROM builder:latest-x86_64_v2, which only exists after build_builder_image.yml has run with this variant. The first master push after merge will fail the compat final leg until the builder workflow is dispatched once to seed that tag. Suggest running build_builder_image.yml manually first.
- Build cost: each builder rebuild now compiles the PETSc/OpenMPI/SLEPc/SUNDIALS chain twice on amd64 (v3 + v2).
- No ARM compat variants: ARMv9 can't be safely built on the GitHub N1 (ARMv8.2) runners, and the existing aarch64 baseline is already the most portable ARM target, so ARM was left as-is.

### DIFF
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -1,0 +1,109 @@
+name: Build image by digest
+description: >-
+  Build a single-platform image, push it to the registry by digest only, and
+  upload the digest as an artifact for a later manifest-merge step. The caller
+  must have checked out the repository first (this is a local action, and the
+  build context is the workspace).
+
+inputs:
+  image:
+    description: Fully-qualified image name (registry/namespace/name).
+    required: true
+  dockerfile:
+    description: Path to the Dockerfile.
+    required: true
+  platform:
+    description: Target platform, e.g. linux/amd64.
+    required: true
+  cache-scope:
+    description: gha cache scope; must be unique per image/variant/arch.
+    required: true
+  artifact-name:
+    description: Name for the uploaded digest artifact; unique per build leg.
+    required: true
+  build-args:
+    description: Newline-separated build-args for build-push-action.
+    required: false
+    default: ""
+  build-secrets:
+    description: Newline-separated secrets for build-push-action.
+    required: false
+    default: ""
+  require-avx512:
+    description: When 'true', fail fast unless the runner CPU has AVX-512.
+    required: false
+    default: "false"
+  registry:
+    description: Registry to log in to.
+    required: true
+  username:
+    description: Registry username.
+    required: true
+  password:
+    description: Registry password/token.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    # x86_64_v4 targets AVX-512, but GitHub's runner fleet is mixed: on a CPU
+    # without it the v4-compiled Spack cmake SIGILLs during configure. Fail fast
+    # here (re-running the job re-rolls the runner) instead of deep in the build.
+    - name: Require AVX-512
+      if: inputs.require-avx512 == 'true'
+      shell: bash
+      run: |
+        if ! grep -qw avx512f /proc/cpuinfo; then
+          echo "::error::Runner CPU lacks AVX-512 (no avx512f in /proc/cpuinfo); this x86_64_v4 image would SIGILL. Re-run to land on an AVX-512-capable runner."
+          exit 1
+        fi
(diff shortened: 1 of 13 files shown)


tbody-cfs:
Removed the `x86_64_v1` variant since this failed to build (https://github.com/boutproject/hermes-3/actions/runs/28881664930/job/85672394345#step:6:550, issue with `readline`). To reenable it, you can just revert [2c983b4](https://github.com/boutproject/hermes-3/pull/608/commits/2c983b46f942ce723d450c9c591378933930a8e7)

dschwoerer:
I am not sure it is worth to do it. For production runs, we want at least v4, which seems to not work in CI, as they are often only v3.
For aarch64, we seem to be stuck at armv8.2, which is also behind armv8.6 for M3 CPUs.

Thus I think it is likely not worth the extra runtime of CI + extra complexity of the CI.

tbody-cfs:
Added a `x86_64_v4` build. This is a bit finicky — not all runners have v4 instructions — so the job auto-restarts up to 5 times until it finds a runner with the right arch

tbody-cfs:
Hmm seems finicky. I close for now

PR #610, opened 2026-07-08
Cleanup tests: Do not download if present

dschwoerer:
### Purpose
Right now the tests always download the data, even if it is present.

This is a waste of resources, and causes problems in restricted environments, where the download is blocked.

This is partially pulled from #577 

### Change Summary
Only download if not present

### Validation
Tests still run in CI, downloading data if needed.
I tested manually that it also works if the direct download is not possible.

### AI Assistance
None

### Documentation
Things should just keep working

### Review Notes

This should be farily limited in scope and easy to review.

### DIFF
--- a/tests/integrated/2D-production/runtest
+++ b/tests/integrated/2D-production/runtest
@@ -41,48 +41,66 @@ expected_filenames = [
 ]
 
 
-## Download files
-tmp_path = zipfile_path.with_name(zipfile_path.name + ".tmp")
+def download_data():
+    ## Download files
+    tmp_path = zipfile_path.with_name(zipfile_path.name + ".tmp")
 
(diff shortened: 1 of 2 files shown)


ZedThree (approved):
It might be nice to move the duplicated code into a common module. Maybe not worth it if it's just the two uses

dschwoerer:
That is probably a good idea, but I am not sure where to move it best?

It would probably good to move also hermes-3 to pytest :innocent:

ZedThree:
With pytest, these could go in a `conftest.py` that gets automatically imported, but even without that, we could just put it in `tests/` and modify `sys.path` in order to import it

mikekryjak:
Thanks @dschwoerer. I think we will definitely have more tests reusing this common test case, so I agree that we need to have a better scheme for this.

PR #611, opened 2026-07-08
Cleanup tests nosymlink

dschwoerer:
### Purpose

Do not symlink to the tests. This makes the setup more fragile, and is not needed.
We can just call hermes-3 directly.

### Change Summary

Remove `ln -s ../../../hermes-3 .` in the tests and call hermes-3 directly.

### Validation

This is part of the tests and thus tested in CI

### AI Assistance

None

### Documentation

This is only an implementation detail, that does not need to be documented.

### DIFF
--- a/tests/integrated/1D-fluid/runtest
+++ b/tests/integrated/1D-fluid/runtest
@@ -2,21 +2,10 @@
 
 # Python script to run and analyse MMS test
 
-from __future__ import division
-from __future__ import print_function
-
-try:
-    from builtins import str
-except:
-    pass
-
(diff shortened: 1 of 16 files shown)


ZedThree (approved):
Switching to pytest would make it very easy to add a fixture with the path to the correct `hermes` executable

mikekryjak:
Thanks @dschwoerer. Here I just aped what Ben was doing in some earlier Hermes-3 tests. I wasn't sure if it was needed, but I didn't know it could cause issues!

PR #612, opened 2026-07-08
Cleanup tests: use mpirun as expected.

dschwoerer:
### Purpose

Previously the arguments to mpirun where modified. This is likely not a good idea.

### Change Summary

mpirun is used as it should be

### Validation

This is covered in CI.

### Documentation

N/A - only a bug fix that caused some strange test failures on some HPC systems.

### Review Notes

If you have a system that does not accept `-np` to specify the number of processors, this should allow to run the tests successfully on that system.

This includes #610 to avoid merge conflicts.

### DIFF
--- a/tests/integrated/2D-production/runtest
+++ b/tests/integrated/2D-production/runtest
@@ -41,48 +41,66 @@ expected_filenames = [
 ]
 
 
-## Download files
-tmp_path = zipfile_path.with_name(zipfile_path.name + ".tmp")
+def download_data():
+    ## Download files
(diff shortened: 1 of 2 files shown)


ZedThree (approved):
It's surprising we need to do this in the first place!

dschwoerer:
I do not think there is a good reason. I suspect this was from an attempt to injects some arguments before `-np` - which is not needed anymore as we switched to env vars, but the previous approach was not fully reverted.

PR #614, opened 2026-07-08
3/zero current changes

malamast:
### Purpose
-Set species momentum to the state and added momentum to outputVars
-Changed the floor to softFloor and the threashold to 10^-7 form 10^-5 

### Question
I am not sure about the permissions. However, it ran just fine for my cases.


cmacmackin (approved):
I've fixed the broken unit tests. and extended them to check for momentum. I also made sure the momentum variable used by `ZeroCurrent::outputVars` is up-to-date.

I'm going to leave it to @mikekryjak to comment on whether the changes to floors are appropriate.

bendudson (approved):

PR #615, opened 2026-07-09
Added floors to components / updated 

malamast:
This is an updated and simplified version of #530


src/braginskii_electron_viscosity.cxx
```diff
@@ -60,10 +60,10 @@ void BraginskiiElectronViscosity::transform_impl(GuardedOptions& state) {
     // SOLPS-style flux limiter
     // Values of alpha ~ 0.5 typically
 
-    const Field3D q_cl = eta * Grad_par(V);   // Collisional value
+    const Field3D q_cl = eta * abs(Grad_par(V)); // Collisional value
```
mikekryjak: `abs(Grad_par(V))` has a discontinuity at 0 which makes it non differentiable there.
A better form is `sqrt(Grad_par(V)^2 + floor^2)` where the floor can be a small number. 

I have applied this method to `abs(Grad_perp(logPnlim))` in `neutral_mixed` and found that floors at `1e-3` and below don't affect the answer. 

I have also done tests with this floor for the velocity gradient for the separated limiters, but only a couple so far, so I don't know what value would work for velocity. You could try with a smaller number like 1e-5 to be safe.

malamast: That's right. I forgot about that. I will fix it but I'll set a smaller number close to the round-off error.

PR #619, opened 2026-07-16
Update BOUT-Spack: fixes python dependency versions for tests.

oparry-ukaea:
### Purpose
Update spack package definitions to pull versions of the python dependencies that are compatible with the current test suite.

### Change Summary
See [BOUT-spack PRs](https://github.com/boutproject/BOUT-spack/pulls?q=is%3Apr+), particularly [this one](https://github.com/boutproject/BOUT-spack/pull/24).

In addition to fixing the python dependencies, these updates cause the hermes and bout packages to build in RelWithDebInfo by default, rather than Release (specify build_type=Release to override).

### Validation
```
. activate_h3env
spack install
spack cd -b hermes-3 && ctest
```
(Checked with both spack 1.1.0 and 1.2.0)

### AI Assistance
None

### DIFF
--- a/spack.yaml
+++ b/spack.yaml
@@ -23,7 +23,7 @@ spack:
   develop:
     boutpp:
       path: ./external/BOUT-dev
-      spec: boutpp@develop
+      spec: boutpp@develop build_type=RelWithDebInfo
     hermes-3:
       path: .
-      spec: hermes-3@develop
+      spec: hermes-3@develop build_type=RelWithDebInfo


cmacmackin (commented):
Why are you are choosing to set the default `build_type` as `RelWithDebug` rather than just specify that setting in `spack.yaml`?

oparry-ukaea:
I wanted spack install to give you the same result as configuring and building with the default cmake options.
But maybe having that be the case only in the dev environment is fine...

oparry-ukaea:
Huh. I thought I got this from CMakeLists, but apparently I imagined it.
BOUT sets Release by default and Hermes doesn't seem to set anything.

cmacmackin:
>I wanted spack install to give you the same result as configuring and building with the default cmake options.
But maybe having that be the case only in the dev environment is fine...

I've seen [reports](https://github.com/spack/spack/issues/22918) that RelWithDebug can actually produce less-optimised builds than Release. Personally prefer not to change the Spack default, especially for BOUT++ if that defaults to Release anyway.

oparry-ukaea:
Ah I knew I'd seen it [somewhere](https://github.com/boutproject/BOUT-dev/blob/9eb94c7c9d9539c1e4e60ec84377a48d5ab50b22/cmake/BuildType.cmake#L12)

But either way, I think you're probably right. I'll revert that change to the packages and set it in spack.yaml instead.

mikekryjak:
That's weird, I thought I heard David say that RelWithDebugInfo is the default.

oparry-ukaea:
Yes, sorry, it is, I was misreading the CMakeLists - it's set in that cmake file linked above

oparry-ukaea:
The hermes-3 and boutpp packages now build in Release by default, but installing them in the dev environment builds them in RelWithDebInfo.

cmacmackin (approved):

oparry-ukaea:
Thanks for the review @cmacmackin!

PR #622, opened 2026-07-20
Add 1D grid file support for fieldline_geometry component

PoloidalLloyd:
### Purpose

Rather than relying on analytical expressions to define geometry such as fieldline radius or poloidal magnetic field, these could be provided exactly from full 2D equilibria and included via a 1D grid file, similar to how goemetry is specified in 2D. This PR adds options to read geometry features from a 1D grid file if present.

### Change Summary

Minor alterations to fieldline_geometry.cxx component, with additions allowing for grid specification of lambda_int, fieldline radius, and poloidal/toroidal field. 

### Validation

Post-processed whether output fieldline_geometry components match input grid file profiles. See example below.

image

### AI Assistance

Yes, minor AI usage to help merge personal fork with newer fieldline_geometry commit on master. Implementation of the grid file methods was done myself, copying other grid file collection methods from the rest of the code base.

### Documentation

TODO: Add documentation on ability to specify geometry with 1D grid file.

### Review Notes

In regards to tests, should we include a separate test to verify if component is correctly matching 1D grid inputs?

### DIFF
--- a/src/fieldline_geometry.cxx
+++ b/src/fieldline_geometry.cxx
@@ -111,29 +111,6 @@ FieldlineGeometry::FieldlineGeometry(std::string name, Options& options, Solver*
     return value;
   };
 
-  // Get the string expressions for lambda_int, fieldline_radius and poloidal_magnetic_field from the options
-  std::string lambda_int_str = geo_options["lambda_int"]
-                                   .doc("Function for the integral heat flux width "
-                                        "lambda_int = lambda_q + 1.64 S [m].")
-                                   .as<std::string>();
-  std::string fieldline_radius_str =
-      geo_options["fieldline_radius"]
-          .doc("Function for the fieldline major radius R [m].")
-          .as<std::string>();
-  std::string poloidal_magnetic_field_str =
-      geo_options["poloidal_magnetic_field"]
-          .doc("Function for the poloidal magnetic field strength Bpol [T].")
-          .as<std::string>();
-
-  // Create FieldGenerator objects for lambda_int, fieldline_radius and poloidal_magnetic_field
-  // These FieldGenerators will be used to evaluate the string expressions
-  FieldGeneratorPtr lambda_int_function =
-      FieldFactory::get()->parse(lambda_int_str, &geo_options);
-  FieldGeneratorPtr poloidal_magnetic_field_function =
-      FieldFactory::get()->parse(poloidal_magnetic_field_str, &geo_options);
-  FieldGeneratorPtr fieldline_radius_function =
-      FieldFactory::get()->parse(fieldline_radius_str, &geo_options);
-
   // Allocate memory for the fields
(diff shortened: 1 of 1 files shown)


src/fieldline_geometry.cxx
```diff
@@ -142,26 +119,72 @@ FieldlineGeometry::FieldlineGeometry(std::string name, Options& options, Solver*
...
-    fieldline_radius[i] = fieldline_radius_function->generate(
-                              bout::generator::Context().set("lpar", lpar[i] * Lnorm))
-                          / Lnorm;
-    poloidal_magnetic_field[i] =
-        poloidal_magnetic_field_function->generate(
-            bout::generator::Context().set("lpar", lpar[i] * Lnorm))
-        / Bnorm;
+  // get lambda_int from grid file or input file
+  if (mesh->sourceHasVar("lambda_int")) {
+    mesh->get(lambda_int, "lambda_int");
```
mikekryjak, line 124: `mesh->get()` only reads domain cells. I'm guessing the guard cells aren't touched and will be zero. Does that cause any problems? I notice that `fieldline_geometry` loops over the whole grid including guards:
https://github.com/boutproject/hermes-3/blob/bdf4d270312522185e657beaa4505164e61f11f7/src/fieldline_geometry.cxx#L135-L138 

I can't think of a reason why we'd need these quantities on the boundary. But there are some operations in `fieldline_geometry` that divide by them, so there could be an issue if some cells are 0:
https://github.com/boutproject/hermes-3/blob/bdf4d270312522185e657beaa4505164e61f11f7/src/fieldline_geometry.cxx#L263

mikekryjak (commented):
Thanks @PoloidalLloyd! I found one possible problem with guard cell handling. 

I think a good unit test would put all these fields in a fake grid file and make sure they exist. Just one test is fine. Here is an example of a unit test that makes a fake mesh and inserts quantities into it: https://github.com/boutproject/hermes-3/blob/a1c3ba381f87686d5a6ee8bf52226886ece6bb1d/tests/unit/test_neutral_full_velocity.cxx#L27-L31

src/fieldline_geometry.cxx
PoloidalLloyd: Sorry this took so long to get around to. Great spot! Do you think its better to add in dummy values in the guard cells (could just pad them with the index to the left or right respectively), or change the loop to "RGN_NOBNDRY"?

mikekryjak: I'm not sure. It depends whether the geometry information is necessary at the boundaries and what it's used for.

dschwoerer: > `mesh->get()` only reads domain cells. 

Are you sure about that? I think it should also read boundary cells, if they are in the mesh file.
I used `mesh->get()` a few times, and always assumed I get the guard cells and ghost cells.

PR #624, opened 2026-07-21
ci: skip docker build workflows outside boutproject/hermes-3

### LINKED ISSUE #623: Docker workflows fail to be skipped on forks
This if statement aims to prevent the Docker actions from running on forks by comparing the root and destination repos in the PR, and skipping the actions if they are not the same:
https://github.com/boutproject/hermes-3/blob/0f44bc7a90a3c6ef37c28e125f101fe484b7c3ce/.github/workflows/test_docker_build.yml#L102-L108

However, on our VANTAGE fork we do a lot of merging between branches of the same repo. This means these workflows still run. My LLM suggests that we simply skip if the source repo is not `boutproject/hermes-3`. @tbody-cfs do you think this is reasonable? I wasn't sure if there was some specific reason that you wrote the check in this way.

tbody-cfs:
### Purpose

Fixes #623. CI workflows were running on downstream forks of `boutproject/hermes-3` when they shouldn't. Two distinct causes, both addressed here:

1. **`pull_request`-triggered docker builds** — the fork guard used `github.event.pull_request.head.repo.full_name == github.repository`, which resolves to the repo the workflow runs in. On a fork, internal branch-to-branch PRs still matched, so the jobs ran and then failed at the publish step.
2. **Scheduled workflows** (raised as a follow-up in the issue) — `build_builder_image`, `cleanup-old-images`, `stale`, and `update-python-test-dependencies` fired on forks via their `schedule` triggers, doing unwanted work (marking a fork's PRs stale, opening dependency PRs) or failing when they tried to publish into the `boutproject` namespace.

### Change Summary

- **`test_docker_build.yml`** — gate the publishing jobs (`build-builder`, `build-final`, `build-jupyter`) on the literal `boutproject/hermes-3` instead of `github.repository`, so they run only where they can publish.
- **Scheduled workflows** — gate each on `github.repository == 'boutproject/hermes-3' || github.event_name == 'workflow_dispatch'`: scheduled/release runs happen only on the canonical repo, but a fork can still manually dispatch.
- **`build_builder_image.yml` (publishing)** — add a `publish_image` toggle to the `build-push-digest` composite action so a fork's manual dispatch builds for validation *without* pushing; the merge/publish jobs are gated to the canonical repo and skip cleanly (via `needs`) on a fork.

| Context | Before | After |
|---|---|---|
| canonical repo, internal PR / schedule / dispatch | runs | runs |
| canonical repo, fork PR | skips | skips |
| downstream fork, scheduled run | runs -> fails | skips |
| downstream fork, manual dispatch | runs -> fails | builds without publishing |

### Validation

Each changed workflow/action parsed with a YAML loader and by the repo's `Validate GitHub Workflows` pre-commit hook. The job-dependency cascade was traced by hand and confirmed with an independent review: on a fork `workflow_dispatch`, `build-builder-image` runs with `publish_image=false`, and `merge-builder-image` -> `build-final-image` -> `merge-final-image` all skip via `needs` rather than fail; on the canonical repo all jobs run normally. Not exercised on a live fork.

### AI Assistance

Changes and this description were drafted with Claude Code. Output was verified by reading the workflow logic, confirming GitHub Actions context/skip semantics (`env` is unavailable in job-level `if`; `inputs` is available in composite-action step `if`), and an independent code-review pass over both commits.

### Documentation

No separate docs needed; the rationale lives in in-file comments next to each guard and in the `publish_image` input description.

### Review Notes

- The `|| github.event_name == 'workflow_dispatch'` clause lets a fork manually run `cleanup-old-images`, `stale`, and `update-python-test-dependencies` too. That is intentional (forks may want to run these against their own repo) and harmless, since those act only on the triggering repo's own resources.
- `boutproject/hermes-3` is left as a literal in the guards (rather than a variable) because `env` is not available in job-level `if:`, and a `vars.*` variable would silently skip everything if unset — the literal is safer and a rename is a simple find/replace.

### DIFF
--- a/.github/actions/build-push-digest/action.yml
+++ b/.github/actions/build-push-digest/action.yml
@@ -28,6 +28,12 @@ inputs:
   token:
     description: Registry password / token (pass secrets.GITHUB_TOKEN).
     required: true
+  push:
+    description: >
+      Whether to push the built image. Set false to build for validation only
+      (e.g. a fork that can't publish); the digest is then not exported, so no
+      merge step should run.
+    default: "true"
   build-args:
     description: Newline-separated build args passed to the build.
     default: ""
@@ -92,9 +98,10 @@ runs:
         cache-to: type=gha,mode=max,scope=${{ inputs.cache-scope }}-${{ steps.vars.outputs.arch }}
         build-args: ${{ inputs.build-args }}
         secrets: ${{ inputs.build-secrets }}
-        outputs: type=image,name=${{ inputs.registry }}/${{ inputs.image-name }},push-by-digest=true,name-canonical=true,push=true
+        outputs: type=image,name=${{ inputs.registry }}/${{ inputs.image-name }},push-by-digest=true,name-canonical=true,push=${{ inputs.push }}
 
     - name: Export digest
+      if: inputs.push == 'true'
       shell: bash
       run: |
         mkdir -p /tmp/digests
@@ -102,6 +109,7 @@ runs:
         touch "/tmp/digests/${digest#sha256:}"
 
     - name: Upload digest
+      if: inputs.push == 'true'
       uses: actions/upload-artifact@v4
       with:
         name: ${{ inputs.cache-scope }}-digests-${{ steps.vars.outputs.arch }}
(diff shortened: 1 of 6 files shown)


mikekryjak (approved):
Thanks @tbody-cfs!

PR #626, opened 2026-07-24
Connection-length flux limiter for conduction

malamast:
WORK IN PROGRESS
More tests are needed. 

## Summary

1) 
-   Added an optional connection-length-based limiter for the parallel  Braginskii heat conductivity in `braginskii_conduction`.
-   The new model follows the free-streaming approximation used in GRILLIX [Zholobenko et al., Nuclear Fusion 64, 106066 (2024)].
-   The existing local, gradient-dependent heat-flux limiter is retained and remains the default, so existing input files should produce the same results as before.
-   The major radius is provided in metres and converted internally to the Hermes-3 length normalization.

2) 
-   Implement the SOLPS Rozhansky et al. (2009) neoclassical correction to the ion parallel thermal conductivity using fixed q95, major radius, and inverse aspect ratio inputs.
-   Reuse the q95 and major-radius options from the connection-length heat-flux limiter and check that all required parameters are positive when either model is enabled.
-   Not implemented for electrons.

## New input options

The limiter model can be selected independently for each species:

```ini
[e]
kappa_limit_model = connection_length
kappa_limit_alpha = 1.0
kappa_limit_q95 = 4.0
kappa_limit_R = 2

[d+]
kappa_limit_model = connection_length
kappa_limit_alpha = 1.0
kappa_limit_q95 = 4.0
kappa_limit_R = 2

### DIFF
--- a/include/braginskii_conduction.hxx
+++ b/include/braginskii_conduction.hxx
@@ -27,7 +27,11 @@ struct BraginskiiConduction : public NamedComponent<BraginskiiConduction> {
   ///   - diagnose                     Output additional diagnostic fields?
   ///   - kappa_coefficient            Heat conduction constant. Default is 3.16 for
   ///                                  electrons, 3.9 otherwise
-  ///   - kappa_limit_alpha            Flux limiter, off by default.
+  ///   - kappa_limit_alpha            Flux limiter, off by default. Free-streaming coefficient.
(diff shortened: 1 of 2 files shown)


mikekryjak:
It would be good to have a flag for these models to only take effect in the core. I think implementing Luciani would also be interesting, but not necessarily in the same PR. Maybe you could make this one flexible and then we can add Luciani later.

PR #627, opened 2026-07-24
Reactions performance improvements

### LINKED ISSUE #625: Too much RHS spent in reaction rate evaluation
Running vtune on `1D-threshold` for one timestep of length 100 reveals something a bit startling. **33% of the runtime is taken up by evaluating AMJUEL polynomials, with a total of 56% of the RHS cost being spent in `Reaction::transform`**. This is even worse for the `annulus-te-he-fixed-neutrals` example, where `Reaction::transform` **costs 67% of the RHS.** Here is the flame graph for 1D-threshold:

Image

I used Claude to perform a detailed report on this issue and compare our reaction evaluation to that in SOLPS, GRILLIX and UEDGE. I used several Opus subagents plus a few Sonnet subagents for verification. The issue is that we are doing many evaluations of what is already not the cheapest reaction form. 

## Easy ways to improve:
- Trick from EIRENE: avoid using `log(n)` and `log(T)` in the evaluation. Precompute them as fields and pass them in.
- Trick from GRILLIX: avoid clamping the density and temperature in the evaluation. Clamp them in the pre-processed fields.
- Compute <sigma v> once for all three channels and all reactants. Currently it's called separately for the density, pressure and momentum channels, and repeated for each reactant:
https://github.com/boutproject/hermes-3/blob/0f44bc7a90a3c6ef37c28e125f101fe484b7c3ce/include/rate_helper.hxx#L192-L202
- Another trick from EIRENE and GRILLIX: in the evaluation kernel, we currently add the polynomials one by one. Apparently this is problematic because of latency reasons. Instead, we can use Horner's rule which does it all in one operation, avoids the use of powers and allows both fused multiply-add and vectorisation:

Image

## More comprehensive fixes on top:
- UEDGE: tabulate the rates to enable 2D linear interpolation. 
- UEDGE: there is a trick that interacts with the colouring stencil where you can save on re-evaluations somehow, but I don't understand this yet. 
- B2.5 style (fluid neutrals): pre-compute a table where the rate is in the form:

$\langle \sigma v \rangle = \exp(\alpha_{0} + \alpha_{1} \ln{T_e})$

where:
$\alpha_1 = \frac{ \partial \ln{\langle \sigma v \rangle} } { \partial \ln{T_e}}$.

The coefficients $\alpha_0$ and $\alpha_1$ contain the density dependence, which is refreshed every time the table is rebuilt (each outer loop). The temperature dependence is there explicitly and has **the additional benefit of having the derivative with respect to temperature, which is used for linearisation.** We could recompute the table at each nonlinear step, or even just at the new timestep. 

Thanks @oparry-ukaea for help with vtune and @jonycastagna for flagging this in his tests.

oparry-ukaea:
### Purpose
Get rid of some inefficiencies in the calculation of reaction rates, particularly from Amjuel data.

Aims to resolve #625.

### Change Summary
- Rejigs `RateHelper::calc_rates()` to 
  1. reduce the number of calls to rate_calc_func (and therefore to the expensive AmjuelData::eval_* functions)
  2. reduce the use of `std::map` in favour of `std::vector`, particularly inside the `BOUT_FOR` loop
- Uses [Horner's method](https://en.wikipedia.org/wiki/Horner%27s_method) for evaluating the bivariate polynomials.

### Validation
I ran the `1D-threshold` example and an abbreviated version of the (2D) `annulus-te-he-fixedneutrals` example in vtune, then examined the hotspot data. Bout++/hermes libs and execs generated with -O3 and -DCHECK=0.

Total time spent in reaction polynomial evaluation (orange), other parts of Reaction::transform (yellow) and the rest of the code (blue)  for `1D-threshold`:
image

and for `annulus-te-he-fixedneutrals`:
image

### AI Assistance
GitHub copilot used for code-completion and assistance with refactoring. 

### Documentation
None; developer interface largely unchanged.

### DIFF
--- a/include/rate_helper.hxx
+++ b/include/rate_helper.hxx
@@ -16,14 +16,24 @@
 
 namespace hermes {
 
-/// Struct to hold pre-averaged data for each cell
-struct CellData {
-  CellData() : CellData(0, 0, 0) {}
-  CellData(BoutReal c, BoutReal l, BoutReal r) : centre(c), left(l), right(r) {}
+/// Struct to hold values used in cell averaging
+struct DataTriple {
+  DataTriple() : DataTriple(0, 0, 0) {}
+  DataTriple(BoutReal c, BoutReal l, BoutReal r) : centre(c), left(l), right(r) {}
   BoutReal centre, left, right;
 };
 
-typedef std::map<std::string, CellData> CellProps;
+// Temporary storage for central,left,right vals of various properties at each cell index
+struct CellProps {
+  CellProps(size_t num_reactants)
+      : densities(num_reactants), collision_freqs(num_reactants) {}
+
+  DataTriple mass_action = {1, 1, 1};
+  DataTriple rate;
+
+  std::vector<DataTriple> densities;
+  std::vector<DataTriple> collision_freqs;
+};
 
 /// Signatures for different rate calculations.
 /// N.B. one extra arg required for the mass action factor.
@@ -31,7 +41,7 @@ using OneDRateFunc = std::function<BoutReal(BoutReal, BoutReal)>;
 using TwoDRateFunc = std::function<BoutReal(BoutReal, BoutReal, BoutReal)>;
 using RateFuncVariant = std::variant<OneDRateFunc, TwoDRateFunc>;
 
-/// Struct to hold reaction rate and collision frequency data
+/// Struct to hold final (possibly averaged) reaction rate and collision frequency fields that get returned by calc_rates()
 struct RateData {
   /// The reaction rate Field3D
   Field3D rate;
@@ -89,7 +99,7 @@ struct RateHelper {
     for (const auto& reactant : reactant_names) {
       // Hold the wrapper in a local so the referenced field lifetime is obvious to GCC.
       const GuardedOptions reactant_density = state["species"][reactant]["density"];
-      this->reactant_densities[reactant] = &reactant_density.GetRef<Field3D>();
+      this->reactant_densities.push_back(&reactant_density.GetRef<Field3D>());
     }
 
     // Compute / extract fields that are required as parameters for the rate calculations
@@ -97,14 +107,14 @@ struct RateHelper {
       static_assert(dependent_false<RateParamsType>,
                     "RateParamsTypes::ET not implemented");
     } else if constexpr (RateParamsType == RateParamsTypes::nT) {
-      for (auto field_lbl : {"e:density", "e:temperature"}) {
-        // Split the access to avoid dangling-reference warnings through temporary wrappers.
-        const GuardedOptions rate_param = state["species"][field_lbl];
-        add_rate_param(field_lbl, rate_param.template GetRef<Field3D>());
-      }
+
+      const GuardedOptions n_e = state["species"]["e:density"];
+      const GuardedOptions T_e = state["species"]["e:temperature"];
+      this->rate_params.push_back(&n_e.GetRef<Field3D>());
+      this->rate_params.push_back(&T_e.GetRef<Field3D>());
     } else if constexpr (RateParamsType == RateParamsTypes::T) {
       calc_Teff(state, units, reactant_names, Teff_storage);
-      add_rate_param("Teff", Teff_storage);
+      this->rate_params.push_back(&Teff_storage);
     } else {
       // Compile-time error if any other RateParamsType enum exists
       static_assert(dependent_false<RateParamsType>, "Unhandled RateParamsType");
@@ -125,28 +135,20 @@ struct RateHelper {
 
     // Initialize result data structure
     RateData result;
-    std::string first_key = str_keys(this->rate_params)[0];
-    result.rate = emptyFrom(*this->rate_params[first_key]);
+    result.rate = emptyFrom(*this->rate_params[0]);
     for (const std::string& reactant : this->reactant_names) {
-      result.collision_frequencies[reactant] = emptyFrom(*this->rate_params[first_key]);
+      result.collision_frequencies[reactant] = emptyFrom(*this->rate_params[0]);
     }
 
-    // Temporary storage for central,left,right vals of each property at each cell index
-    CellProps cell_props;
-    cell_props["rate"] = CellData();
-    for (const std::string& reactant : this->reactant_names) {
-      // CellData for the collision frequency associated with each reactant
-      cell_props[reactant] = CellData();
-      // CellData for the density associated with each reactant
-      cell_props[reactant + "_density"] = CellData();
-    }
+    // Temporary storage for densities, rate and collision frequencies associated with a cell
+    CellProps cell_props(this->num_reactants);
 
     // Populate cell_props differently according to the rate function type
     std::visit(
         [this, &cell_props, &do_averaging, &result](auto&& rate_calc_func) {
           auto J = result.rate.getCoordinates()->J;
           BOUT_FOR(i, region) {
-            // Add densities and mass action-factor to cell_props
+            // Get densities and mass action-factor for this cell
             add_densities_and_mass_action(i, cell_props, do_averaging);
 
             using RateFuncType = std::decay_t<decltype(rate_calc_func)>;
@@ -154,16 +156,16 @@ struct RateHelper {
               // 1D rate function
               if constexpr (RateParamsType == RateParamsTypes::T) {
                 // Get rate params
-                CellData T_data = compute_rate_param(Teff_name, i, do_averaging);
+                DataTriple T_data = compute_rate_param(0, i, do_averaging);
 
                 // Compute rate(s)
-                cell_props["rate"].centre =
-                    rate_calc_func(cell_props["mass_action"].centre, T_data.centre);
+                cell_props.rate.centre =
+                    rate_calc_func(cell_props.mass_action.centre, T_data.centre);
                 if (do_averaging) {
-                  cell_props["rate"].left =
-                      rate_calc_func(cell_props["mass_action"].left, T_data.left);
-                  cell_props["rate"].right =
-                      rate_calc_func(cell_props["mass_action"].right, T_data.right);
+                  cell_props.rate.left =
+                      rate_calc_func(cell_props.mass_action.left, T_data.left);
+                  cell_props.rate.right =
+                      rate_calc_func(cell_props.mass_action.right, T_data.right);
                 }
               } else {
                 throw BoutException(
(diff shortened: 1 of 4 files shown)


oparry-ukaea:
Marking this ready-for-review as the existing changes are relatively undisruptive and already give a big perfomance boost.
I propose we use a new PR for further improvements.

include/rate_helper.hxx
```diff
@@ -212,26 +188,37 @@ struct RateHelper {
...
               }
             } else {
               // Extract rate, collision frequencies at cell centre and store in result
-              result.rate[i] = cell_data["rate"].centre;
-              for (const std::string& reactant : this->reactant_names) {
-                result.collision_frequencies[reactant][i] = cell_data[reactant].centre;
+              result.rate[i] = cell_props.rate.centre;
+              for (std::size_t reactant_idx = 0; reactant_idx < this->num_reactants;
+                   reactant_idx++) {
+                std::string reactant_name = this->reactant_names[reactant_idx];
```
cmacmackin:
```suggestion
                const std::string& reactant_name = this->reactant_names[reactant_idx];
```

include/rate_helper.hxx
```diff
@@ -212,26 +188,37 @@ struct RateHelper {
...
-                    4. / 6 * cell_data[reactant].centre
-                    + (Ji + Jm) / (12. * Ji) * cell_data[reactant].left
-                    + (Ji + Jp) / (12. * Ji) * cell_data[reactant].right;
+              result.rate[i] = 4. / 6 * cell_props.rate.centre
+                               + (Ji + Jm) / (12. * Ji) * cell_props.rate.left
+                               + (Ji + Jp) / (12. * Ji) * cell_props.rate.right;
+              for (std::size_t reactant_idx = 0; reactant_idx < this->num_reactants;
+                   reactant_idx++) {
+
+                std::string reactant_name = this->reactant_names[reactant_idx];
```
cmacmackin:
```suggestion
                const std::string& reactant_name = this->reactant_names[reactant_idx];
```

include/rate_helper.hxx
```diff
@@ -123,83 +131,51 @@ struct RateHelper {
...
                     "Unhandled RateParamsType (1D rate function being passed)");
               }
             } else if constexpr (std::is_same_v<RateFuncType, TwoDRateFunc>) {
+              // 2D rate function
               if constexpr (RateParamsType == RateParamsTypes::nT) {
-                CellData ne_data = compute_rate_param("e:density", i, do_averaging);
-                CellData Te_data = compute_rate_param("e:temperature", i, do_averaging);
+                // Get rate params
+                DataTriple ne_data = compute_rate_param(0, i, do_averaging);
+                DataTriple Te_data = compute_rate_param(1, i, do_averaging);
```
cmacmackin: It would be clearer if we defined named constants for these indices.

include/rate_helper.hxx
```diff
@@ -311,59 +288,57 @@ private:
...
+   * @param do_averaging[in] whether to compute left and right values for averaging
    */
-  inline CellData compute_collision_freq_density(Ind3D i, bool do_averaging,
-                                                 const std::string& reactant_name,
-                                                 const CellData& mass_actions) {
-    auto ym = i.ym();
-    auto yp = i.yp();
-    const auto& dens = *this->reactant_densities.at(reactant_name);
-    CellData result = mass_actions;
+  inline void add_density_factors(CellProps& cell_props, bool do_averaging = true) {
```
cmacmackin: Given this involves multiplication, calling this `add_density_factors` is confusing. Maybe call it `apply_density_factors` instead.

include/rate_helper.hxx
```diff
@@ -311,59 +288,57 @@ private:
...
-   * @param do_averaging whether to compute left and right values
-   * @return CellData struct containing centre, left, and right mass action factors
-   */
-  inline CellData compute_mass_actions(Ind3D i, bool do_averaging) {
-    auto ym = i.ym();
-    auto yp = i.yp();
-    CellData result{1, 1, 1};
-    for (const auto& [sp_name, dens] : this->reactant_densities) {
-      result.centre *= (*dens)[i];
+  inline void add_densities(Ind3D i, CellProps& cell_props, bool do_averaging) {
```
cmacmackin: Again, there really isn't any addition going on here. Maybe call it `collect_densities`?

include/rate_helper.hxx
```diff
@@ -16,20 +16,30 @@
...
-using TwoDRateFunc = std::function<BoutReal(BoutReal, BoutReal, BoutReal)>;
+// Temporary storage for central,left,right vals of various properties at each cell index
+struct CellProps {
+  CellProps(size_t num_reactants)
+      : densities(num_reactants), collision_freqs(num_reactants) {}
+
+  DataTriple rate;
+
+  std::vector<DataTriple> densities;
+  std::vector<DataTriple> collision_freqs;
```
cmacmackin, line 34: Something to consider for future optimisation: it might be better to have multiple vectors (e.g., `density_left`, `density_centre`, `density_right`) in place of a vector of structs. This reduces caches misses in the cases where you aren't actually accessing all three values held in `DataTriple` (as is the case in `add_density_factors` when `do_averaging == false`). You'd probably need to do some profiling to determine whether that would be helpful in this case though.

include/rate_helper.hxx
```diff
@@ -123,83 +131,51 @@ struct RateHelper {
...
+            // Get densities for this cell
+            add_densities(i, cell_props, do_averaging);
+
             using RateFuncType = std::decay_t<decltype(rate_calc_func)>;
             if constexpr (std::is_same_v<RateFuncType, OneDRateFunc>) {
+              // 1D rate function
               if constexpr (RateParamsType == RateParamsTypes::T) {
-                CellData T_data = compute_rate_param(Teff_name, i, do_averaging);
+                // Get rate params
+                DataTriple T_data = compute_rate_param(0, i, do_averaging);
```
cmacmackin: It would be clearer if we defined a named constants for this index.

src/amjuel_data.cxx
```diff
@@ -24,27 +24,26 @@ namespace hermes {
...
+    BoutReal row_result = 0.0;
+    for (std::size_t icol = ncols; icol-- > 0;) {
+      row_result = row_result * logntilde + coeff_table[irow][icol];
     }
-    logT_n *= logT;
+    result = result * logT + row_result;
   }
-  return exp(result) * 1e-6; // Note: convert cm^3 to m^3
+
+  return exp(result) * 1e-6;
```
cmacmackin, line 47: Why did you remove this comment?

cmacmackin (approved):
I think this looks good, aside from some minor quibbles. I didn't follow all of the changes that were made though (as this is a part of the code with which I'm unfamiliar). Someone that understand the physics better should probably review it as well. It looks like it's quite well covered by unit tests, so that gives me some confidence at least.

include/rate_helper.hxx
oparry-ukaea: Changed as suggested.

include/rate_helper.hxx
oparry-ukaea: Changed as suggested.

include/rate_helper.hxx
oparry-ukaea: Done

include/rate_helper.hxx
oparry-ukaea: 'add' was supposed to mean 'add to cell_props' but agree that wasn't very clear - renamed as suggested.

include/rate_helper.hxx
oparry-ukaea: renamed as suggested

include/rate_helper.hxx
oparry-ukaea: Good point, thanks. I'll give that a try in the next round of optimisations.

include/rate_helper.hxx
oparry-ukaea: Done

src/amjuel_data.cxx
oparry-ukaea: Oops, meant to move it to the line above - fixed.

include/rate_helper.hxx
```diff
@@ -212,26 +191,37 @@ struct RateHelper {
...
-                    + (Ji + Jm) / (12. * Ji) * cell_data[reactant].left
-                    + (Ji + Jp) / (12. * Ji) * cell_data[reactant].right;
+              result.rate[i] = 4. / 6 * cell_props.rate.centre
+                               + (Ji + Jm) / (12. * Ji) * cell_props.rate.left
+                               + (Ji + Jp) / (12. * Ji) * cell_props.rate.right;
+              for (std::size_t reactant_idx = 0; reactant_idx < this->num_reactants;
+                   reactant_idx++) {
+
+                const std::string& reactant_name = this->reactant_names[reactant_idx];
+                result.collision_frequencies[reactant_name][i] =
```
mikekryjak: Claude suggests that you could try to remove this `map` as well.

src/amjuel_data.cxx
```diff
@@ -12,8 +12,8 @@ namespace hermes {
 /// @brief Helper functions for evaluating Amjuel fits.
 
 /**
- * @brief Evaluate an Amjuel double polynomial fit in n and T, given a table of
- * coefficients (see page 20 of amjuel.pdf).
+ * @brief Evaluate an Amjuel double polynomial fit in n and T.
+ * This function user's Honer's method rather than the formulation on page 20 of amjuel.pdf.
```
mikekryjak:
```suggestion
 * This function user's Horner's method rather than the formulation on page 20 of amjuel.pdf.
```

mikekryjak (changes requested):
Looks good to me apart from some minor things. Have you checked if anything in the docs needs changing?

include/rate_helper.hxx
oparry-ukaea: I was resisting this, because I was imagining that using vectors in the results struct would complicate/uglify data access once it's passed back to Reaction::tranform_impl. But thinking about it, I could just hide the indices behind an interface in RateData. I'll give it a go.

src/amjuel_data.cxx
oparry-ukaea: Fixed

oparry-ukaea:
> Looks good to me apart from some minor things. Have you checked if anything in the docs needs changing?

Thanks for the reminder - reaction developer docs updated too.

mikekryjak (approved):
Happy if @cmacmackin is happy.

oparry-ukaea:
@cmacmackin - are you happy for me to merge this?

cmacmackin:
Yes, looks good to me!

PR #628, opened 2026-07-27
Ion viscosity permissions

malamast:
WIP 

1) Fixed permissions. We need AA when bounce_frequency is enabled.
2) Apply parallel BC for eta. We probably don't need the values of eta at the parallel guard cells.  It depends on how the parallel operators are implemented. 
3) Applied a softFloor to collision frequency 1/nu as in head_conduction.


src/braginskii_ion_viscosity.cxx
```diff
@@ -282,6 +283,37 @@ void BraginskiiIonViscosity::transform_impl(GuardedOptions& state) {
       eta.getMesh()->communicate(eta);
       eta.applyBoundary("neumann");
 
+      // Note: We need parallel boundary conditions on eta
+      if (eta.hasParallelSlices()) {
```
mikekryjak, line 287: I don't understand parallel slices well, but Claude tells me that this if statement is always going to be true since `eta.getMesh()->communicate(eta)` always calculates the parallel slices. 

A similar BC in `braginskii_conduction` has a simpler implementation: https://github.com/boutproject/hermes-3/blob/a1c3ba381f87686d5a6ee8bf52226886ece6bb1d/src/braginskii_conduction.cxx#L276-L289

mikekryjak:
Hi @malamast! Thanks for this. I'm not an expert on viscosity but found a possible issue and left a comment. I have two other suggestions:

- Can you change the name of the PR so that it mentions the fact this changes the ion viscosity parallel boundary condition? I think that is probably the most important change in this PR. 
- Now that eta is `neumann`, it is non-zero at the boundary, which means viscosity will now flow through the parallel boundary. Is this intended? If so, could you mention the reason in the PR description? 
- In `braginskii_conduction` we set kappa to Neumann but then disable conduction through the boundary by passing a bool to `Div_par_k_Grad`:

https://github.com/boutproject/hermes-3/blob/a1c3ba381f87686d5a6ee8bf52226886ece6bb1d/src/braginskii_conduction.cxx#L294-L295

src/braginskii_ion_viscosity.cxx
```diff
@@ -441,6 +473,15 @@ void BraginskiiIonViscosity::outputVars(Options& state) {
                       {"long_name", std::string("nu star") + species_name},
                       {"species", species_name},
                       {"source", "ion_viscosity"}});
+
+      set_with_attrs(
+          state[std::string("eta_") + species_name], d.eta,
+          {{"time_dimension", "t"},
+           {"units", "Pa s"},
+           {"conversion", Pnorm / Omega_ci},
+           {"long_name", std::string("Ion viscosity coefficient") + species_name},
```
mikekryjak, line 482: For consistency with conduction:
```suggestion
           {"long_name", species_name + std::string(" ion viscosity coefficient")},
```

Issue #469, opened 2026-01-14
Need performance tests


mikekryjak:
We aren't controlling for performance regressions right now. 

One way to do this would be to use `aispeed-velocity`: https://asv.readthedocs.io/en/stable/using.html
This tool can capture the performance of your tests and gives you a nice performance dashboard. 

The setup is allegedly straightforward, although you need to set up a separate repo where it will commit the performance results.

This has been implemented in a different repo by an RSE here at UKAEA: https://github.com/ukaea/powerbalance/tree/main/benchmarks

mikekryjak:
That's quite cool! However, I think we need more for Hermes-3, because performance is an absolutely critical aspect of the code. I think being able to track performance history would be helpful.

PR #630, opened 2026-07-31
Add performance benchmarks

cmacmackin:
### Purpose
As discussed in #469, we need to be able to track whether changes to Hermes-3 result in performance regressions. This PR adds a basic [airspeed velocity](https://asv.readthedocs.io/en/latest/index.html) setup to perform benchmarking.

### Change Summary
Configurations have been added to run a couple representative cases with Airspeed Velocity (asv). To do this I needed a way to build and install the package into a Python environment. I created a `pyproject.toml` file with a [builder that wraps cmake](https://scikit-build-core.readthedocs.io/en/stable/index.html) to create a Python wheel. There are some kludges, to ensure the hermes-3 executable gets installed in the `bin` directory (rather than a subdirectory of `site-packages`). Ideally we would write this a bit better and maybe integrate it with the machinery that builds wheels for BOUT++. Possibly it could become one of the ways we distribute hermes-3 in future. I also refactored the 2D-production test slightly so I could more easily reuse its routines for downloading test data. This was all done in a bit of a rush and possibly could use some tidying up.

**Warning:** A gotcha that confused me for a while is that asv only benchmarks changes that have been *committed* to your repo (at least by default). Changes that have not been committed will not be run.

### Validation
I have confirmed that ASV can run the benchmarks successfully. I have confirmed that `2D-production` still runs after my refactoring.

### AI Assistance
None.

### Documentation
There are not yet any documentation updates. It would probably be useful to add some instructions to the developer documentation.

### Review Notes
There are still tasks left to complete on this, but I'm going on holiday for two weeks and wanted to share my progress so far before that. These task are

- [x] Create a CI workflow that checks whether a PR creates any performance regressions compared to `master`. This can be done using the [asv continuous](https://asv.readthedocs.io/en/latest/commands.html#asv-continuous) command and/or the [freegs workflow](https://github.com/freegs-plasma/freegs/blob/master/.github/workflows/benchmark.yml) written for this purpose.
- [ ] Develop a workflow for long-term tracking of hermes-3 performance. There are [other repositories](https://github.com/newton-physics/newton-asv/tree/main) with [workflows](https://github.com/newton-physics/newton/blob/main/.github/workflows/aws_gpu_benchmarks.yml) that provide examples of how this can be done. This will require a dedicated server, however, so will be happening on a slightly longer time-horizon.
- [ ] Add a 3D simulation for benchmarking.
- [x] Tweak the number and size of timesteps for the benchmark cases.
- [x] Consider whether it is worth running the benchmarks multiple times for better statistics.
- [ ] Possibly tidy up the Python wheel a bit and create workflows that will build and deploy these for each release.
- [x] Add developer documentation on running benchmarks.

If reviewers would like to implement any of these while I am away they are welcome to do so. Otherwise I can look at them upon my return.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -44,6 +44,7 @@ jobs:
       OMPI_MCA_rmaps_base_oversubscribe: yes
       MPIRUN: mpiexec -np
       LD_LIBRARY_PATH: /home/runner/local/lib:$LD_LIBRARY_PATH
+      ASV_FACTOR: 1.1
 
     steps:
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -88,7 +88,7 @@ jobs:
         run: |
           # ID of this machine
           uv run --no-project asv machine --yes
-          base_sha="$(git show -s --format='%H' ${{ github.event.pull_request.base.label }})"
+          base_sha="$(git show -s --format='%H' origin/${{ github.base_ref }})"
 
           echo ""
           echo ""
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,63 @@
+# Run the asv benchmarks on this PR, comparing to the base commit
+# Adapted from scikit-image and freegs:
+# https://github.com/scikit-image/scikit-image/blob/main/.github/workflows/benchmarks.yml
+# https://github.com/freegs-plasma/freegs/blob/master/.github/workflows/benchmark.yml
+
+name: Benchmarks
+
+on: pull_request
+
+env:
+  PR_HEAD_LABEL: ${{ github.event.pull_request.head.label }}
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+
+    steps:
+      # We need the full repo
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements_bench.txt
+
+      - name: Run benchmarks
+        env:
+          ASV_FACTOR: 1.5
+        run: |
+          # ID of this machine
+          asv machine --yes
+
+          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Contender: ${GITHUB_SHA} ($PR_HEAD_LABEL)"
+
+          # Run benchmarks for current commit against base
+          ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
+          asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
+              | sed "/Traceback \|failed$\|PERFORMANCE DECREASED/ s/^/::error::/" \
+              | tee benchmarks.log
+
+          # Report and export results for subsequent steps
+          asv compare --split --factor $ASV_FACTOR ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA}
+          if grep "Traceback \|failed\|PERFORMANCE DECREASED" benchmarks.log > /dev/null ; then
+              exit 1
+          fi
+
+      - name: Add instructions to artifact
+        if: always()
+        run: cp benchmarks.log .asv/results/
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: asv-benchmark-results-${{ runner.os }}
+          path: .asv/results
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -5,7 +5,31 @@
 
 name: Benchmarks
 
-on: pull_request
+on:
+  pull_request:
+    paths:
+      - '**,cxx'
+      - '**.hxx'
+      - '**.cmake'
+      - '**.in'
+      - 'CMakeLists.txt'
+      - 'external/BOUT-dev'
+      - 'pyproject.toml'
+      - 'benchmarks/**.py'
+      - 'asv.conf.json'
+      - 'tests/integrated/1D-recycling/data/BOUT.inp'
+      - 'tests/integrated/2D-production/data/BOUT.inp'
+      - 'examples/tokamak-2D-driftplane/2D-drift-plane-turbulence-te-ti/BOUT.inp'
+      - 'tests/integrated/runtest_utils.py'
+      - '.github/workflows/benchmarks.yml'
+      - '!performance/'
+      - '!tests/unit/'
+
+# One run per PR/ref; cancel superseded runs since only the latest
+# push matters
+concurrency:
+   group: performance-benchmark-${{ github.ref }}
+   cancel-in-progress: true
 
 env:
   PR_HEAD_LABEL: ${{ github.event.pull_request.head.label }}
@@ -60,14 +84,12 @@ jobs:
         run: uv sync --only-dev
 
       - name: Run benchmarks
-        env:
-          ASV_FACTOR: 1.5
         run: |
           # ID of this machine
           uv run asv machine --yes
 
-          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
-          echo "Contender: ${GITHUB_SHA} ($PR_HEAD_LABEL)"
+          echo "\n\nBaseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Contender: ${GITHUB_SHA} (merge of $PR_HEAD_LABEL into ${{ github.event.pull_request.base.label }})\n"
 
           # Run benchmarks for current commit against base
           ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -14,7 +14,8 @@ jobs:
   benchmark:
     runs-on: ubuntu-latest
     env:
-      OMP_NUM_THREADS: ${{ matrix.config.omp_num_threads }}
+      OMP_NUM_THREADS: 1
+      OPENBLAS_NUM_THREADS: 1
       PYTHONPATH: ${{ github.workspace }}/tools/pylib
       OMPI_MCA_rmaps_base_oversubscribe: yes
       MPIRUN: mpiexec -np
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -87,19 +87,22 @@ jobs:
       - name: Run benchmarks
         run: |
           # ID of this machine
-          uv run asv machine --yes
+          uv run --no-project asv machine --yes
 
-          echo -e "\n\nBaseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
-          echo -e "Contender: ${GITHUB_SHA} (merge of $PR_HEAD_LABEL into ${{ github.event.pull_request.base.label }})\n"
+          echo ""
+          echo ""
+          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Contender: ${GITHUB_SHA} (merge of $PR_HEAD_LABEL into ${{ github.event.pull_request.base.label }})"
+          echo ""
 
           # Run benchmarks for current commit against base
           ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
-          uv run asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
+          uv run --no-project asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
               | sed "/Traceback \|failed *$\|PERFORMANCE DECREASED/ s/^/::error::/" \
               | tee benchmarks.log
 
           # Report and export results for subsequent steps
-          uv run asv compare --split --factor $ASV_FACTOR ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA}
+          uv run --no-project asv compare --split --factor $ASV_FACTOR ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA}
           if grep "Traceback \|failed *$\|PERFORMANCE DECREASED" benchmarks.log > /dev/null ; then
               exit 1
           fi
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -89,8 +89,8 @@ jobs:
           # ID of this machine
           uv run asv machine --yes
 
-          echo "\n\nBaseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
-          echo "Contender: ${GITHUB_SHA} (merge of $PR_HEAD_LABEL into ${{ github.event.pull_request.base.label }})\n"
+          echo -e "\n\nBaseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo -e "Contender: ${GITHUB_SHA} (merge of $PR_HEAD_LABEL into ${{ github.event.pull_request.base.label }})\n"
 
           # Run benchmarks for current commit against base
           ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -88,16 +88,17 @@ jobs:
         run: |
           # ID of this machine
           uv run --no-project asv machine --yes
+          base_sha="$(git show -s --format='%H' ${{ github.event.pull_request.base.label }})"
 
           echo ""
           echo ""
-          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Baseline:  ${base_sha} (${{ github.event.pull_request.base.label }})"
           echo "Contender: ${GITHUB_SHA} (merge of $PR_HEAD_LABEL into ${{ github.event.pull_request.base.label }})"
           echo ""
 
           # Run benchmarks for current commit against base
           ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
-          uv run --no-project asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
+          uv run --no-project asv continuous $ASV_OPTIONS ${base_sha} ${GITHUB_SHA} \
               | sed "/Traceback \|failed *$\|PERFORMANCE DECREASED/ s/^/::error::/" \
               | tee benchmarks.log
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -95,12 +95,12 @@ jobs:
           # Run benchmarks for current commit against base
           ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
           uv run asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
-              | sed "/Traceback \|failed$\|PERFORMANCE DECREASED/ s/^/::error::/" \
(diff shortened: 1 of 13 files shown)


benchmarks/benchmarks.py
```diff
@@ -0,0 +1,83 @@
...
+        os.chdir(self.runpath)
+        # Load the runtest file to access methods needed to download data
+        sys.path.append(str(testdir))
+        # print(testdir / "runtest")
+        # spec = importlib.util.spec_from_file_location("runtest", testdir / "runtest")
+        # assert spec is not None
+        # assert spec.loader is not None
+        # runtest = importlib.util.module_from_spec(spec)
+        # sys.modules["runtest"] = runtest
+        # spec.loader.exec_module(runtest)
```
ZedThree: Some commented out code:
```suggestion
```

tests/integrated/2D-production/runtest_utils.py
```diff
@@ -0,0 +1,76 @@
...
+    "BOUT.restart.5.nc",
+    "BOUT.restart.6.nc",
+    "BOUT.restart.7.nc",
+    "BOUT.restart.8.nc",
+    "BOUT.restart.9.nc",
+    "grid_test2_allpump.nc",
+]
+
+
+class DataDownloader:
```
ZedThree: If we're moving this, maybe put it in `tests/integrated`? Then we can (separately) update the recycling test to use this?

ZedThree (commented):
Looks good. It might be useful as some preliminary work to run this over a bunch of commits to get a feel for how much the measure performance varies?

Is the intention to add CI in this PR?

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def setup(self):
+        self.rundir = tempfile.TemporaryDirectory()
+        self.runpath = pathlib.Path(self.rundir.name)
+        testdir = TESTPATH / "2D-production"
+        shutil.copytree(testdir / "data", self.runpath / "data")
+        self.cwd = os.getcwd()
+        os.chdir(self.runpath)
+        # Load the runtest file to access methods needed to download data
+        sys.path.append(str(testdir))
```
cmacmackin: I need to pop this from the path in the teardown method (if not before).

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,83 @@
...
+    timeout = 180
+    rounds = 1
+    repeat = 1
+    number = 1
+    warmup_time = 0
+
+    def setup(self):
+        self.rundir = tempfile.TemporaryDirectory()
+        self.runpath = pathlib.Path(self.rundir.name)
+        # FIXME: is this the right test? My notes say 1D-thene
```
cmacmackin:
```suggestion
```

cmacmackin:
>Looks good. It might be useful as some preliminary work to run this over a bunch of commits to get a feel for how much the measure performance varies?

Unfortunately I don't think that will be easy to do. ASV won't be able to install any commits that don't already have the Python wheel configuration set up. Possibly it won't be able to run at all on commits that don't have an ASV config; I'm not sure.

>Is the intention to add CI in this PR?

I hadn't really made a decision on that but I'd be inclined to do so unless @mikekryjak is in a rush to get this merged.

ZedThree:
I am reasonably certain that ASV uses the config and actual benchmark tests from the current commit when running on past commits -- but it has been several years since I've used it, so I could be mistaken.

If that is the case, we could just put the cmake build line in [`build_command`](https://asv.readthedocs.io/en/stable/asv.conf.json.html#build-command-install-command-uninstall-command) instead of using `scikit-build-core`?

To be clear, I don't think this is a blocker, it just might be a useful activity.

cmacmackin:
The reason I didn't want to just use cmake in `build_command` is it wouldn't have played very nicely with ASV's ability to cache previous builds (as far as I could tell).

cmacmackin:
The benchmark test is failing because the base commit of this PR doesn't support building the Python wheel for hermes-3. Our options are:

- Merge it anyway. We know the reason for the failure and that it won't arise again once this has been merged.
- Split out the Python wheel changes into a separate PR and merge that first. Then rebase this commit. That way the wheel can be built for both the base commit of the PR.
- Scrap the whole idea of building Python wheels and configure asv to call CMake directly. This would allow it to run on any commit, but would make it difficult to take adavantage of its caching of old builds. It will only work if there are *no* changes required to other files to get them to play nicely with asv.

dschwoerer:
> > Looks good. It might be useful as some preliminary work to run this over a bunch of commits to get a feel for how much the measure performance varies?
> 
> Unfortunately I don't think that will be easy to do. ASV won't be able to install any commits that don't already have the Python wheel configuration set up. Possibly it won't be able to run at all on commits that don't have an ASV config; I'm not sure.

Would it be a lot of work to do that only temporarily, to do some testing?
I do think we should aim for the better long-term solution, and that is likely caching builds.

The failing CI is not an issue, I think. If it would break, it needs to be fixed "quickly" - but we only merge a commit every few days, so having a new CI failure until that is fixed would not be that bad.

tests/integrated/2D-production/runtest_utils.py
cmacmackin: Done.

cmacmackin:
@ZedThree I've now addressed all of your comments (apart from making `asv` use cmake directly for the build, which I think is more trouble than it's worth). I've also created a CI workflow to check for performance regressions. The only thing left which we might want to include in this PR is we've asked @bendudson if there is a good 3D case for benchmarking, which he is looking into. I guess if he has something before you and @mikekryjak approve this PR then I'll add it. Otherwise we'll merge this and put in a separate PR for a 3D test case later.

cmacmackin:
> Would it be a lot of work to do that only temporarily, to do some testing?

I honestly don't know how hard it would be. @mikekryjak has already manually checked for performance regressions in key commits over approximately the last year and not found anything, so he was inclined not to bother with this.

ZedThree (approved):
Looks great, thanks @cmacmackin!

ZedThree:
I guess one thing to be aware of if we haven't done the historic testing is the allowable jitter may need some fine tuning, we'll have to wait and see!

cmacmackin:
> I guess one thing to be aware of if we haven't done the historic testing is the allowable jitter may need some fine tuning, we'll have to wait and see!

We can keep an eye on this, but there is a tolerance of 10% which I think should be plenty.

dschwoerer:
Just out of curiosity, would it also be possible to add a different compile configuration, so that we could run a FCI run and check there for performance regressions? It is certainly out of scope for this PR ...

mikekryjak:
> > Would it be a lot of work to do that only temporarily, to do some testing?
> 
> I honestly don't know how hard it would be. @mikekryjak has already manually checked for performance regressions in key commits over approximately the last year and not found anything, so he was inclined not to bother with this.

Yes, I have built an agent-assisted way to do historical bisection easily. I can do this separately.

cmacmackin:
>Just out of curiosity, would it also be possible to add a different compile configuration, so that we could run a FCI run and check there for performance regressions? It is certainly out of scope for this PR ...

I think it would be pretty difficult. As far as I'm aware, `asv` is designed to work with only one set of build/install options (it's meant for Python, which doesn't have the same sort of build options as you tend to get in a compiled project).

tests/integrated/2D-production/runtest
```diff
@@ -23,84 +21,21 @@ data_dir = this_dir / "data"
...
-            )
-
-        # Copy bits of the file from response to a temp file
-        # This ensures no partial files are left if the download fails
-        with open(tmp_path, "wb") as out_file:
-            shutil.copyfileobj(response, out_file)
+sys.path.append(str(this_dir.parent))
+from runtest_utils import DataDownloader2DProduction
+
+dd = DataDownloader2DProduction("2D-Prodcution", this_dir)
```
mikekryjak:
```suggestion
dd = DataDownloader2DProduction("2D-Production", this_dir)
```

tests/integrated/2D-recycling/runtest
```diff
@@ -32,107 +30,44 @@ this_dir = Path(__file__).parent
...
+        t_download = time()
 
-    extractZip()
+    dd.extractZip()
 
     t_extract = time()
     if verbose:
         print(
-            f"2D-recycling test: zip extraction took {t_extract - t_download:.2f} seconds"
+            f"2D-Production test: zip extraction took {t_extract - t_download:.2f} seconds"
```
mikekryjak:
```suggestion
            f"2D-recycling test: zip extraction took {t_extract - t_download:.2f} seconds"
```

tests/integrated/2D-recycling/runtest
```diff
@@ -32,107 +30,44 @@ this_dir = Path(__file__).parent
...
 if not debug:
-    t_download = time()
-    if not checkHash(False):
-        download_data()
-        checkHash()
+    if not dd.checkHash(False):
+        dd.download_data()
+        t_download = time()
+        if verbose:
+            print(f"2D-Production test: download took {t_download - t0:.2f} seconds")
```
mikekryjak:
```suggestion
            print(f"2D-recycling test: download took {t_download - t0:.2f} seconds")
```

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,92 @@
...
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Install dev tools
+        run: uv sync --only-dev
+
+      - name: Run benchmarks
+        env:
+          ASV_FACTOR: 1.5
```
mikekryjak: If this is the same as --factor here https://asv.readthedocs.io/en/stable/commands.html then the test fails if the slowdown is 1.5, and not 1.1 like stated in the docs.

asv.conf.json
```diff
@@ -0,0 +1,204 @@
...
+
+    // FIXME: How best to run this in CI?
+
+    // Customizable commands for building the project.
+    // See asv.conf.json documentation.
+    // To build the package using pyproject.toml (PEP518), uncomment the following lines
+    "build_command": [
+        "python -m pip install build",
+        "python -m build",
+        "python -m pip wheel -w {build_cache_dir} {build_dir}"
```
mikekryjak, line 25: From my friend Claude:

"python -m pip install build",
"python -m build",
"python -m pip wheel -w {build_cache_dir} {build_dir}"

python -m build builds an sdist and a wheel into {build_dir}/dist, then pip wheel compiles the whole thing again from scratch in a separate isolated env — nothing is reused. For a Python project that costs seconds; for Hermes-3 + BOUT++ it is a full C++ compile each time, times two commits. Dropping the python -m build line (and the pip install build) should roughly halve the 55-minute job. It is in asv's template, but the template assumes a cheap build.

Related: "install_timeout": 1200 (https://github.com/boutproject/hermes-3/blob/c4291b8a6ab38f1972e133dbf6fe8dfb56e4917e/asv.conf.json#L70) is 20 minutes per build command. With the double build on a 4-core runner that is uncomfortably close to the edge; a timeout surfaces as a benchmark failure and fails the PR.

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,77 @@
...
+    testpath: pathlib.Path = DEFAULT_TESTPATH
+    datadir: str = "data"
+    data_downloader: type[AbstractDataDownloader] | None = None
+
+    def __repr__(self) -> str:
+        return self.testname
+
+
+class Simulations:
+    timeout = 180
```
mikekryjak: Just to confirm - your local tests were oversubscribed and took about 40s right?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,92 @@
...
+
+          # Run benchmarks for current commit against base
+          ASV_OPTIONS="--split --show-stderr --factor $ASV_FACTOR"
+          uv run asv continuous $ASV_OPTIONS ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA} \
+              | sed "/Traceback \|failed$\|PERFORMANCE DECREASED/ s/^/::error::/" \
+              | tee benchmarks.log
+
+          # Report and export results for subsequent steps
+          uv run asv compare --split --factor $ASV_FACTOR ${{ github.event.pull_request.base.sha }} ${GITHUB_SHA}
+          if grep "Traceback \|failed\|PERFORMANCE DECREASED" benchmarks.log > /dev/null ; then
```
mikekryjak:
```suggestion
          if grep "Traceback \|failed$\|PERFORMANCE DECREASED" benchmarks.log > /dev/null ; then
```

tests/integrated/runtest_utils.py
```diff
@@ -0,0 +1,81 @@
+import hashlib
+import shutil
+import urllib.request
+import zipfile
+from pathlib import Path
+from typing import ClassVar
+
+
+class AbstractDataDownloader:
```
mikekryjak, line 9: Does this find a way to avoid redownloading the zip in every test? Note it's the same files for both tests

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,92 @@
+# Run the asv benchmarks on this PR, comparing to the base commit
+# Adapted from scikit-image and freegs:
+# https://github.com/scikit-image/scikit-image/blob/main/.github/workflows/benchmarks.yml
+# https://github.com/freegs-plasma/freegs/blob/master/.github/workflows/benchmark.yml
+
+name: Benchmarks
+
+on: pull_request
```
mikekryjak: Since these are quite expensive, maybe we could add an exception for docs and formatting commits using `paths-ignore`. We could also make sure there's no duplicated runs like in https://github.com/boutproject/hermes-3/blob/2a79cf58268ce4378bedc8e0419b1057d861513c/.github/workflows/test_docker_build.yml#L37-L40.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Run benchmarks
+        env:
+          ASV_FACTOR: 1.5
+        run: |
+          # ID of this machine
+          uv run asv machine --yes
+
+          echo "Baseline:  ${{ github.event.pull_request.base.sha }} (${{ github.event.pull_request.base.label }})"
+          echo "Contender: ${GITHUB_SHA} ($PR_HEAD_LABEL)"
```
mikekryjak: From Claude:

▎ On pull_request events GITHUB_SHA is the auto-generated merge commit, not the PR head — so this labels it with PR_HEAD_LABEL, but the hash is on no branch of that name.
▎
▎ I'd keep benchmarking the merge commit; it's the better choice, since head.sha on a stale branch would blame unrelated master changes on the PR. Just make the label honest:
▎ echo "Contender: ${GITHUB_SHA} (merge of `$PR_HEAD_LABEL` into ${{ github.event.pull_request.base.label }})"
▎ One consequence worth knowing: .asv/results is keyed by that hash, which GitHub regenerates whenever the base branch moves. Fine for the pass/fail gate, but the uploaded artifact can't be traced to a real commit later — which matters for the long-term tracking on your task list. Logging head.sha alongside would cover it.

mikekryjak (changes requested):
Thanks @cmacmackin. I went through it, then used Claude to go through it again and find more things. I checked its comments apart from the ones I quoted verbatim.

.github/workflows/benchmarks.yml
cmacmackin: Good catch

.github/workflows/benchmarks.yml
cmacmackin: I've decided to do this with `paths` rather than `ignore-paths`, but it will accomplish the same thing. It just makes it easier to ignore changes to tests.

benchmarks/benchmarks.py
cmacmackin: Local tests were *not* oversubscribed. Looking at the [CI logs](https://github.com/boutproject/hermes-3/actions/runs/32381259050/job/96464751324?pr=630#step:7:4597), I see that 2D-production takes about 2.8 minutes (168 seconds) so I think it would be worth increasing this limit to 300 seconds.

tests/integrated/runtest_utils.py
cmacmackin: I wasn't thinking about that problem, so no. It wouldn't be that difficult to put in a cache directory, or similar, but I think that is beyond the scope of this PR.

asv.conf.json
cmacmackin: This is strange. I think the asv developers have made a mistake here. Commands I used were the ones they had put in the default `asv.conf.json` file for `pyproject.toml` builds. True, repeating the build of the wheel is trivial if something is a pure Python project, but there is still no good reason to do it. Furthermore, the asv developers clearly do imagine asv being used by people writing Python modules in compiled languages. I think it is worth raising an issue about this.

cmacmackin: I've doubled the install timeout.

.github/workflows/benchmarks.yml
cmacmackin: Do we expect the word "failed" to always be at the end of a line?

cmacmackin: Actually, I just deliberately induced failures and here is the output:
```
[100.00%] ··· ======================= ========
                      testcase                
              ----------------------- --------
                 1D-recycling-dthe     failed 
                   2D-production       failed 
               tokamak-2D-driftplane   failed 
              ======================= ========
```
If you look closely there is a space after "failed", so we don't want to make this change. So long as no one names a testcase something including the word `failed` it shouldn't cause any problems.

.github/workflows/benchmarks.yml
cmacmackin: >I'd keep benchmarking the merge commit; it's the better choice, since head.sha on a stale branch would blame unrelated master changes on the PR.

I don't follow this logic. If we're working on a stale commit and changes to `master` have resulted in a slow-down, then testing the merged branch would fail (because they'd be comparing against a very old commit on `master`) even if the changes in the PR haven't introduced any slowdown.

It is the standard behaviour in GitHub for PR CI workflows to run on the merged version of the code so I'm perfectly happy to keep doing that here and to update the output to be more explicit about it. I just really don't understand that particular sentence.

.github/workflows/benchmarks.yml
mikekryjak: the `sed` line has `failed$`, so they should be consistent:
https://github.com/boutproject/hermes-3/pull/630/changes/BASE..c4291b8a6ab38f1972e133dbf6fe8dfb56e4917e#diff-22eef62c9d7b36d02f5fd5aaada92702e22d6795c464239cb7b8fe0f26ea1e1cR75

I was thinking that `failed$` would be better in case the word could come up during e.g. solver messages. But if it doesn't come up now, then leaving an over-eager checker which flags any time there is a new instance of `failed` is not a problem. So maybe we should make both `failed` then.

tests/integrated/runtest_utils.py
mikekryjak: Ok. Honestly, I am getting tired of updating the Zenodo and I'm finding that it's very easy to make mistakes in doing this. With PETSc on CI, we could run a simulation from scratch in a reasonable timeframe instead and only download the grids. We could consolidate it then.

.github/workflows/benchmarks.yml
mikekryjak: I think the suggestion is that the workflow is still run on the merged version of the code, but that the performance is compared against the merge commit's first parent (on the master side), instead of the base of the PR, i.e. the first commit of the branch.

The wording wasn't very clear which is partly why I didn't quite understand it either (and hoped that it would make more sense to you).

.github/workflows/benchmarks.yml
cmacmackin: I've changed both regexes so they now read `Traceback \|failed *$\|PERFORMANCE DECREASED`. That way it will allow whitespace between `failed` and the end of the line.

.github/workflows/benchmarks.yml
cmacmackin: That's not how I read what Claude had said. Regardless, I've checked and the way the CI is currently configured will compare the merge commit with the `master` commit from which this branch diverged. I think Claude may just have been talking gobbledygook here.

cmacmackin:
OK, everything appears to be working except that `asv` iisn't able to build the baseline commit inthe benchmark CI because there isn't the configuration to build the `wheel` in that commit. To be honest, I don't know how this CI workflow was passing for a while. As [previously discussed](https://github.com/boutproject/hermes-3/pull/630#issuecomment-5342762300), I say we override the failing workflow and merge.

.github/workflows/benchmarks.yml
mikekryjak: Yeah, from what I understand this is exactly the problem. We shouldn't test against the `master` commit from which this branch diverged, since `master` could have evolved since then. We should test it against the latest `master` commit which we will be merging into.

cmacmackin: OK, I could change to benchmark against the HEAD of `master`. Just to be clear though, I don't think that's at all what Claude was saying in the quote you sent me.

mikekryjak (approved):
Thanks @cmacmackin!

PR #636, opened 2026-08-12
Fix some inconsistencies in `sheath_boundary` with super sonic flow

yichenfu:
### Purpose
#574 introduced supersonic flow in `sheath_boundary`, but the implementation is inconsistent when the `phi` component is not calculated. This PR intends to correct this issue.

Specifically, the ion speed at the sheath boundary is defined [here](https://github.com/boutproject/hermes-3/blob/c9d74b69b08979be2e6dc0efb720a120998695f0/src/sheath_boundary.cxx#L858) as follows, which allows supersonic flow speed.
```C++
const BoutReal visheath = std::max(sqrt(C_i_sq), // Positive -> into sheath
                                   Vi[i]);       // Allow supersonic flow

// Set boundary conditions on flows
Vi[ip] = 2. * visheath - Vi[i];
```

However, later when calculating ion charge flux`ion_sum` [here](https://github.com/boutproject/hermes-3/blob/c9d74b69b08979be2e6dc0efb720a120998695f0/src/sheath_boundary.cxx#L284), the formula still uses the sound speed `C_i_sq` instead of `visheath`:
```C++
ion_sum[i] += s_i * Zi * sin_alpha[ip] * sqrt(C_i_sq);
```
So, when the flow is supersonic, the calculation of `ion_sum` is inconsistent. This will happen when the `phi` component is undefined, such as in cases with no `vorticity` component. 

Similar behavior also exists when the secondary electron emission coefficient is nonzero (`ion_ee_gamma_max`>0). The calculation of `ion_flux_over_ne` [here](https://github.com/boutproject/hermes-3/blob/c9d74b69b08979be2e6dc0efb720a120998695f0/src/sheath_boundary.cxx#L379) also used an inconsistent ion speed:
```C++
const BoutReal ion_flux_over_ne = s_i * sin_alpha[ip] * sqrt(C_i_sq); // Γ_i / n_e
```

### Change Summary
The calculation of `visheath` is introduced before calculating `ion_sum` and `ion_flux_over_ne`, which is then used to calculate the charge flux.

### Validation
I tested a case with `sheth_boundary` but with no `vorticity`; it runs well.

### AI Assistance
Opus 5 is used to help diagnose the issue and verify solutions.

### DIFF
--- a/src/sheath_boundary.cxx
+++ b/src/sheath_boundary.cxx
@@ -208,6 +208,7 @@ void SheathBoundary::transform_impl(GuardedOptions& state) {
     struct IonBoundaryInfo {
       Field3D density;
       Field3D temperature;
+      Field3D velocity;
       BoutReal mass;
       BoutReal charge;
       BoutReal adiabatic;
@@ -215,7 +216,8 @@ void SheathBoundary::transform_impl(GuardedOptions& state) {
 
     std::vector<IonBoundaryInfo> ion_species;
 
-    // Need to sum  s_i Z_i C_i over all ion species
+    // Sum the positive outgoing ion charge flux divided by electron density:
+    // s_i * Z_i * sin(alpha) * |V_i,sheath|
     //
     // To avoid looking up species for every grid point, this
     // loops over the boundaries once per species.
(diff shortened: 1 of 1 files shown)


dschwoerer (approved):
Looks good. I have not verified that all places are now modified, that should be, but it certainly does improve things.

One note about test coverage: All lines are covered but the unit tests clearly do not cover the supersonic case. It would be nice to have a test for this, especially if @ZedThree is going to update #382

bendudson (approved):
Thanks @yichenfu ! Sorry for the delay.

PR #642, opened 2026-08-26
Add automatic sources when species densities are low

oparry-ukaea:
### Purpose
Adds an optional LowNSources component that artificially boosts species densities to keep them away from their respective floors.

### Change Summary
`LowNSources` is the orchestrator component, constructing instances of `PseudoReaction`automatically based on the species that are enabled in the simulation
The 'standard' strategy (no others implemented yet) uses the following logic:
- If atoms are in short supply, but ions aren't (i.e. ions are at or above their threshold), pseudo-recombination boosts the neutral density
- If ions are in short supply, but atoms aren't, pseudo-ionisation boosts the ion density
- If both atoms and ions are in short supply, an external source turns on and increases in strength for larger ion deficits, such that there are also enough ions to allow pseudo-recombination
- If molecules are in short supply, pseudo-association boosts the molecular density

The main two classes are:

1. `PseudoReaction`
  - Instances are independent of the existing physical `Reaction` subclasses, but mimic several real processes. They can be of type 'ionization', 'recombination', 'association' or 'external_src'
  - The existing code in `Reaction::transform_impl()` adds source terms, but the usual json-based rates are bypassed - custom rates are provided via the overridable `Reaction::get_rates()`
  - PseudoReaction rates (may) depend on
     - a 'target' species - the source activates when the density of that species falls below its threshold (fixed_factor * its density_floor)
    - an optional 'boost' species - the source increases in strength the further that species falls below its threshold
    - an optional 'gate' species - the source can be disabled entirely if the gate species falls below its threshold
    - an option to turn the source off if the boost species is below its threshold

2. `LowNSources`
  - Parses species list from the config on construction and decides which pseudo-reactions are needed.
  - Instantiates and configures PseudoReaction instances automatically
  - When transform() is called, loops over all pseudo-reactions calling their transform functions on the state.
  - Similar loop when outputVars() is called, which allows pseudo reactions to write out their diagnostics (i.e the source sizes) 

---

ToDo:
- [ ] Unify the functionality provided by the `SpeciesType` and `SpeciesTypeAlt` enums and related code.
- [ ] Simplify (or preferably remove the need for entirely) the dummy reaction data class used by PseudoReaction components.
- [ ] Rename some low-n member vars
- [ ] Add some user documentation, possibly dev guidance too
- [ ] Implement or remove the 'uedge' strategy for setting source rates
- [ ] Investigate noisy source behaviour near thresholds in integration test
- [ ] Some kind of performance benchmark against the existing code. 

---

### Validation
There are unit tests to check that
- the correct pseudo-reactions are constructed for a variety of different species combinations
- source sizes are exactly zero if the target species is above its threshold
- (ion) source sizes get larger the further the density falls below its threshold
- source sizes are as expected when the ion density is high (no external source of ions, particle number conserved)
- source sizes are as expected when the ion density is low (includes external ion source effects)
- corresponding momentum and energy sources are also  added with the expected sizes

and an integration test that
- initialises some atoms and ions with densities mostly below their respective and checks that the sources turn on and off as expected. A plotting script is also provided; resulting animation should look like this:

low-n_sources_animation

---

### AI Assistance
LLMs used for in-line code completion and refactoring.

### Documentation
ToDo

---

### Review Notes
- Does the `LowNSourcesConstructionTest.ConstructorBuildsOrderedPairsByElement` unit test include representative examples for all of the likely species combinations we'll need?
- The integration test shows some noisy behaviour where the atomic density is close to its threshold. Not clear to me whether this is a problem - things get smoothed out again fairly quickly after moving away from the (possibly pathological) initial conditions.


src/species_parser.cxx
```diff
@@ -73,39 +102,39 @@ SpeciesParser::SpeciesParser(const std::string& species_str) {
...
+  this->species_str = construct_species_str(this->base_species, this->charge);
 }
 
 ///
-SpeciesParser::SpeciesParser(const std::string element, int charge)
-    : element(element), charge(charge) {
-  this->species_str = construct_species_str(element, charge);
+SpeciesParser::SpeciesParser(std::string base_species, int charge)
+    : base_species(base_species), charge(charge) {
+  this->species_str = construct_species_str(base_species, charge);
```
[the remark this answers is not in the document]
oparry-ukaea: Fixed this instead by modifying the secondary constructor functionality to become copy+change charge, which saves parsing the species string again unnecessarily.

src/low_n_sources.cxx
```diff
@@ -0,0 +1,439 @@
...
+  for (auto& coll_freq : result.collision_frequencies) {
+    coll_freq = zeroFrom(n_target);
+  }
+
+  // Set rate such that the source term will restore the target species to its threshold density over delta_t = [low_n_timescale]
+  //   (deficit = source = 0  when n_target > n_thresh)
+  BOUT_FOR(i, region) {
+    const BoutReal n_deficit =
+        std::max<BoutReal>(0.0, this->density_threshold - n_target[i]);
+    result.rate[i] = n_deficit / this->low_n_timescale;
```
totork, line 384: I am curious, do you think this choice of setting this "hard" threshold might cause trouble for implicit solvers because it might have problems when the solution does fluctuate around the threshold?

oparry-ukaea: It might.
I think I need a more physically plausible example (than the integration test) in order to do a meaningful comparison against the current approach, but definitely worth checking.

PR #644, opened 2026-08-26
Adjust Braginskii conduction for Fci

totork:
### Purpose
Adjust the Braginskii_conduction component to be functional in Fci. 

### Change Summary

- Not clearing parallel slices when Fci
- Use different boundary approach for kappa_par
- Add MMS test

### Validation
1D Slab MMS test for electron heat conduction

### AI Assistance
None

### Documentation
None

### Review Notes
I am not sure if mesh->communicate(kappa) is needed here to be honest.

### DIFF
--- a/src/braginskii_conduction.cxx
+++ b/src/braginskii_conduction.cxx
@@ -236,7 +236,9 @@ void BraginskiiConduction::transform_impl(GuardedOptions& state) {
     // EvolvePressure::finally
 
     Field3D P = GET_VALUE(Field3D, species["pressure"]);
-    P.clearParallelSlices();
+    if (!P.isFci()) {
+      P.clearParallelSlices();
+    }
     const Field3D Pfloor = floor(P, 0.0); // Restricted to never go below zero
     const Field3D T = get<Field3D>(species["temperature"]);
     const Field3D N = get<Field3D>(species["density"]);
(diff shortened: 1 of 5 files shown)


src/braginskii_conduction.cxx
```diff
@@ -276,18 +279,26 @@ void BraginskiiConduction::transform_impl(GuardedOptions& state) {
...
 
-    for (RangeIterator r = mesh->iterateBndryLowerY(); !r.isDone(); r++) {
-      for (int jz = 0; jz < mesh->LocalNz; jz++) {
-        auto i = indexAt(kappa_par, r.ind, mesh->ystart, jz);
-        auto im = i.ym();
-        kappa_par[im] = kappa_par[i];
+    // Fci does not work with mesh->iterateBndryLowerY(), so set the boundaries differently
+    if (P.isFci()) {
+
+      mesh->communicate(kappa_par);
```
dschwoerer, line 285: You mean whether we need this? `Div_par_K_Grad_par_mod` I think needs to `Field3DParallel`, so we need parallel slices. Looking at the code above, we need this. For the non-FCI branch, it is also need, as `Grad_par(T)` does not give y boundaries and halos, so this is needed, too.

totork: My idea was in the direction of also calculating the slices of kappa_par instead of communicating because all the quantities would have parallel slices. But e.g. the collision times do not have parallel slices. So they would probably be wrong/inaccurate.

PR #647, opened 2026-08-28
Add more sheath boundary tests; fix some small bugs

ZedThree:
### Purpose
Expand test coverage for the various sheath boundaries and fix some minor bugs

### Change Summary
Adds more unit tests for `sheath-boundary-simple`, plus some basic integrated tests for `sheath-boundary/-simple/-insulating`. These are all just golden answer/regression tests.

### Validation
This just tests current behaviour.

### AI Assistance
No AI was used.

### Documentation
N/A

### Review Notes
Fixed bugs:
- wrong index in calculating `vesheath` in one boundary in `sheath-boundary-insulating`
- sign error (and wrong long name) in calculation of particle source in `sheath-boundary-simple`
- when `phi` isn't set but required in `sheath-boundary-simple`, it could be wrongly calculated in unaligned coordinates


src/sheath_boundary_simple.cxx
```diff
@@ -619,7 +614,7 @@ void SheathBoundarySimple::transform_impl(GuardedOptions& state) {
           BoutReal power = heatflow / dv; // [Wm^-3]
           ASSERT2(std::isfinite(power));
           energy_source[i] += power; // Note: Sign negative because power > 0
-          particle_source[i] -=
+          particle_source[i] +=
```
dschwoerer, line 617: vi is negative, so the whole rhs should be negative. If it is really the source, than that should be fine.

src/sheath_boundary_simple.cxx
```diff
@@ -619,7 +614,7 @@ void SheathBoundarySimple::transform_impl(GuardedOptions& state) {
           BoutReal power = heatflow / dv; // [Wm^-3]
           ASSERT2(std::isfinite(power));
           energy_source[i] += power; // Note: Sign negative because power > 0
```
dschwoerer, line 616:
```suggestion
          energy_source[i] -= power; // Note: Sign negative because power > 0
```
Or does the comment need fixing? `power` seems to be positive, thus the "source" from the sheath should be negative?

tests/integrated/1D-sheath-insulating/runtest
```diff
@@ -0,0 +1,174 @@
...
+success &= check_case(
+    "Sheath higher Ge",
+    expected,
+    "sheath_boundary_insulating:secondary_electron_coef=0.99",
+)
+
+################################################################################
+# Re-run with flipped direction. Should be mirror of base case
+
+flipped_upstream = base_downstream
```
dschwoerer:
```suggestion
flipped_upstream = base_downstream.copy()
```
To not accidentally modify the base values?
I know that they are not used anymore, but it could lead to hard to find bugs.

dschwoerer (commented):
Looks good. might be nice to add a unit test for particle and energy source from the sheath, to make sure they have the expected sign.

src/sheath_boundary_simple.cxx
ZedThree: I'm pretty sure it's the comment that needs fixing, because the energy source should be identical when we flip which end the sheath boundary is at, and that's what we currently see.

The comment is identical between the two boundaries, so most likely a copy-paste error (and the same reason `particle_source` needed fixing?)

src/sheath_boundary_simple.cxx
ZedThree: I think this fix is correct -- the particle source should be symmetric when we flip which end has the sheath boundary. It should also be negative, as particles leave the boundary, balancing the flow into the boundary.

dschwoerer (approved):

src/sheath_boundary_simple.cxx
mikekryjak: This seems likely. Power is definitely positive because the sign difference is handled in the calculation of `q`.

src/sheath_boundary_simple.cxx
mikekryjak: I think the fix is correct. The particle source should be negative as this is the loss of ions into the sheath (maybe it should be better named).

src/sheath_boundary_simple.cxx
```diff
@@ -278,8 +274,7 @@ void SheathBoundarySimple::transform_impl(GuardedOptions& state) {
       }
     }
 
-    phi.allocate();
-    phi.setDirectionY(YDirectionType::Aligned);
+    phi = emptyFrom(Ne);
```
mikekryjak, line 277: @ZedThree for my benefit, could you explain what this change does?

tests/integrated/1D-sheath-insulating/runtest
```diff
@@ -0,0 +1,174 @@
...
+success &= check_case(
+    "Sheath higher Ge",
+    expected,
+    "sheath_boundary_insulating:secondary_electron_coef=0.99",
+)
+
+################################################################################
+# Re-run with flipped direction. Should be mirror of base case
+
+flipped_upstream = base_downstream.copy()
```
mikekryjak, line 156: If the `.copy()` is necessary for this test, then what about the other tests for consistency?

tests/integrated/1D-sheath-simple/runtest
```diff
@@ -0,0 +1,230 @@
...
+}
+
+success &= check_case("Mirrored BCs", expected, "lower_sheath=true")
+
+################################################################################
+# Linear-free boundary
+
+expected = {
+    "upstream": base_upstream,
+    "downstream": {
```
mikekryjak, line 204: Can you include functionality to generate the golden answer for all the tests please?

tests/integrated/1D-sheath-insulating/runtest
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python3
```
mikekryjak, line 1: Can you please add even a very brief description of the tests to the docs under https://hermes3.readthedocs.io/en/latest/developer.html#tests?

mikekryjak (changes requested):
Thanks @ZedThree, some changes requested and comments below.

src/sheath_boundary_simple.cxx
ZedThree: For reference, the `emptyFrom` docstring:

```cpp
/// Return an empty shell field of some type derived from Field, with metadata
/// copied and a data array that is allocated but not initialised.
```

I forget the exact reason why this change was necessary, I assume I found that there are some circumstances where we don't actually want Y to be `Aligned`, but we do always want it to be "like" `Ne`.

Usually when we have this pattern:

```cpp
Field3D function(const Field3D& input) {
  Field3D result;
  result.allocate();
  ...
}
```

we most likely want `result = emptyFrom(input)` (or `result = zeroFrom(input)`)

tests/integrated/1D-sheath-insulating/runtest
ZedThree: It's only necessary here because we're modifying the values

PR #650, opened 2026-09-02
Separate neutral flux limiters

mikekryjak:
### Purpose

This PR makes the flux limiters for diffusive neutral processes in `neutral_mixed` more accurate. It is the final implementation of the separate limiters I worked on in https://github.com/boutproject/hermes-3/pull/556, https://github.com/boutproject/hermes-3/pull/451, https://github.com/mikekryjak/hermes-3/pull/15 and https://github.com/mikekryjak/hermes-3/pull/6. 

Previously, conduction and viscosity were calculated from the already limited diffusion coefficient. However, the calculation of free streaming heat and momentum flux is different than free streaming particle flux, and so viscosity and conduction need a separate limiter calculation. In addition, viscosity and conduction have parallel components which must be limited according to the parallel gradients.

This PR separates the flux limiters into:
- Cross-field advection
- Cross-field conduction
- Cross-field viscosity
- Parallel conduction
- Parallel viscosity

The separated limiters introduce more a lot more nonlinearity in the equations and can lead to degraded performance and stability. Because of this, legacy behaviour with combined limiters remains enabled by default. To switch to the separate limiters, set:
```
[d]
combined_limiters = false
```

### Change Summary

- [x] Separate limiter calculation for `Dnn`, `kappa_n_par`, `kappa_n_perp`, `eta_n_par`, `eta_n_perp`. 
- [x] Limiting is controlled by the free-streaming fraction for each coefficient, or by an explicit desired maximum diffusion coefficient value. Both the free-streaming fraction and explicit limit can be set simultaneously. In this case, the resulting cap is a harmonic average of the two. 
- [x] The regularisation from https://github.com/boutproject/hermes-3/pull/605 was extended to the new limiters. Due to the gradient in `eta_n_max` being a shear rate instead of a scale length, it gets a separate setting for the floors and ceilings.
- [x] Consider deriving the viscosity regularisation from the diffusion/conduction regularisation (perhaps by using thermal speed to convert the units).
- [x] Additional diagnostics for the diffusion coefficients and their limiters.
- [x] Previously, all diffusion coefficients were Dirichlet(0) by default, and evolved variables were Neumann. This doubles up on preventing boundary flux and is confusing. Now, all diffusion coefficients are now Neumann by default, so that the user just needs to change the radial BCs to allow outflows, and contour plots with guard cells showing will look less confusing. 
- [x] Conductivity and viscosity were not zero at the sheath in previous incarnations of separate limiters. Now they are correctly zeroed at the boundary if the `sheath_yup`/`sheath_ydown` flags are enabled.
- [x] Tidy up and improved consistency of radial/parallel BC setting.
- [x] Refactor limiter application, so that diagnostics of `Dmax`, `kappa_max`, `eta_max` etc will contain the effects of both the free-streaming and explicit limit caps. 
- [x] Add unit tests and docs.
- [ ] Add the plasma perpendicular velocity component.

### Validation

I have compared the results against the historic branches which were used to compare to SOLPS (see top of the message).
The existing integrated tests are unchanged because this PR preserves legacy behaviour by default.
New unit tests were added.

### AI Assistance

LLM agents were used extensively for planning and checking the work. I wrote all code, commits and Github content myself.

### Documentation

I have written a new section on `neutral_mixed` boundary conditions. I am still extending the docs in the equations section. 

### Review Notes

This PR changes the structure of `neutral_mixed` quite a bit. I would welcome comments on whether it's a good balance of best C++ practice and readability.


src/neutral_mixed.cxx
```diff
@@ -135,10 +198,24 @@ NeutralMixed::NeutralMixed(const std::string& name, Options& alloptions, Solver*
...
+                              "<0 means off")
+                         .withDefault(-1.0)
+                     / (Nnorm * meters * meters / seconds);
+
+  viscosity_limit =
+      options["viscosity_limit"]
+          .doc("Upper limit on dynamic viscosity eta_n [Pa s] = [kg m^-1 s^-1]. "
+               "<0 means off")
+          .withDefault(-1.0)
+      / (SI::Mp * Nnorm * meters * meters / seconds);
```
mikekryjak, line 218: Docs units inconsistent with normalisation?

PR #658, opened 2026-09-18
Topological sort debugging

bendudson:
### Purpose
Adds capabilities and instructions on how to disable and debug component ordering. 

### Change Summary
A setting `hermes:topological_sort_components` can be set to `false` to turn off topological sorting (printing a warning). An integer `hermes:debug_component_order` will print component order and permissions if greater than zero.

### Validation
Unit tests check that the ordering is performed and disabled. Other unit tests check that debug messages are printed.

### AI Assistance
Codex/gpt-5.4 used to plan and implement changes.

### Documentation
Added a guide on debugging component order issues.


bendudson:
Not needed because #445 already exists.

